Issue #16, opened 2019-12-20
create a custom markup for inline blocks


shcheklein:
We need to write an HTML snippet to embed a link.

Can we utilize mdx or create a custom MD markup to do this?

iAdramelk:
Depending on the syntax that we want, there is multiple ways to achieve it:

1. Use [remark-custom-blocks](https://github.com/zestedesavoir/zmarkdown/tree/master/packages/remark-custom-blocks) to add simple syntax blocks.
2. Use pseudo-html tags with [react-rehype](https://using-remark.gatsbyjs.org/custom-components/)
3. Use actual MDX.

Can you write examples of the snippets that we need to create with all fields that these snippets have?

iAdramelk:
Or we can just transform markdown AST in any way we want.

shcheklein:
Yep, you can check it here (or almost in any other blog post):

https://github.com/iterative/blog/edit/master/content/blog/2019-12-14-november-19-dvc-heartbeat.md

```html
<a href="https://www.deploymachinelearning.com/" class="external-link-preview">
  <section class="elp-content-holder">
    <div class="elp-description-holder">
      <h4 class="elp-title">Deploy Machine Learning Models with Django</h4>
      <div class="elp-description">Version 1.0 (04/11/2019) Piotr Płoński The demand for Machine Learning (ML) applications is growing. Many resources…</div>
      <div class="elp-link">deploymachinelearning.com</div>
    </div>
    <div class="elp-image-holder">
      <img src="/uploads/images/2019-12-14/deploy-machine-learning-models.png" />
    </div>
  </section>
</a>
```

shcheklein:
@iAdramelk considering your three options, could you convert that example above into different representations so that we can pick the simplest one?

iAdramelk:
We have four fields in the snippet above:

- *href*: "https://www.deploymachinelearning.com/"
- *title*: "Deploy Machine Learning Models with Django"
- *description*: "Version 1.0 (04/11/2019) Piotr Płoński The demand for Machine Learning (ML) applications is growing. Many resources..."
- *picture*: "/uploads/images/2019-12-14/deploy-machine-learning-models.png"

This number of fields excludes custom blocks, but other two options remains.

We can do with that something like:

```
<external-link
  href="https://www.deploymachinelearning.com/"
  title="Deploy Machine Learning Models with Django"
[...]

PR #53, opened 2020-01-05
Fix: create a custom markup for inline blocks

oleksandrbyk:
@shcheklein 
#16  is done on this PR.
I utilized MDX to do this.


shcheklein:
@oleksandrbyk looks like deployment is failing

oleksandrbyk:
@shcheklein yeah, let me fix it.

src/components/markdown/index.tsx
```diff
@@ -1,17 +1,17 @@
...
 
-function Markdown({ html }: IMarkdownProps) {
+function Markdown({ body }: IMarkdownProps) {
   return (
-    <div
-      className={styles.wrapper}
-      dangerouslySetInnerHTML={{ __html: html }}
-    />
+    <div className={styles.wrapper}>
+      <MDXRenderer>{body}</MDXRenderer>
```
shcheklein: so, does it worth moving the processing to the client? what is the performance-wise impact of this change? can we implement this server-side purely?

src/components/nav/index.tsx
```diff
@@ -11,6 +11,7 @@ function Nav() {
   return (
     <div className={styles.wrapper}>
       <div className={styles.links}>
+        // @ts-ignore
```
shcheklein: could explain please why wasn't it failing before? is it a proper fix for this?

oleksandrbyk: @shcheklein 
Just ran `yarn typecheck`.
I think the argument current is always equal to `undefined`, so for now, I added a ts-ignore.
We can remove the argument as it always refers to `undefined`.

oleksandrbyk:
@shcheklein now, the deployment failed is fixed.

src/components/markdown/index.tsx
oleksandrbyk: Yeah. I am also concerned about badly affecting the performance of this change, but it would be little trifle. Yeah, it is possible and I am now trying to implement this server-side purely.

PR #55, opened 2020-01-05
Update - create a custom markup for inline blocks #16

sekmet:
Hello,
I got done the milestone requested here: https://github.com/iterative/blog/issues/16. 
It was updated to MDX markdown engine and all gatsby plugins are working.

Following the requirements the external links snippets can be used in markdown files as describe bellow:

-------------------------------------------

< ExternalLink
  href="https://www.deploymachinelearning.com/"
  title="Deploy Machine Learning Models with Django"
  description="Version 1.0 (04/11/2019) Piotr Płoński The demand for Machine Learning (ML) applications is growing. Many resources..."
  picture="/uploads/images/2019-12-14/deploy-machine-learning-models.png"
/ >

------------------------------------------

Please check the updated DVC blog live here: 
https://inspiring-brown-042ba4.netlify.com/

I love your project, Thank you!
Best regards
Carlos Guimaraes


shcheklein:
@sekmet it looks good, but let's make it to the point. 1. no need to change CMS (it will be removed, I think we have a ticket for this) 2. no style changes that are not related to this PR (disregard Restyled) 3. no code changes unrelated to this PR. 4. is it possible to pre-render HTML end-to-end always on the client side w/o the need to render on the client?

sekmet:
@shcheklein Hello :), about the points: 1. Ok, no problem. Will you guys use another one, just curious? 2. Ok, the style change was applied to fix remark-embed-gist plugin style issue under mdx/js engine. Theres another way fix this: Adding 'gatsby-remark-embed-gist/gatsby-ssr' to gatsby-config. but for some reason yet to find, not works always. 3. Sure, no problem. 4. I think yes, its possible using a skeleton template approach.In this case a pre-rendered page displays a skeleton template while the data waits to be rehydrated with React hydration function, this deliver a super-fast experience to the users. No re-render on page transitions.

shcheklein:
@sekmet 

> Will you guys use another one, just curious?

I hope not. I don't like them. I don't like to support the code around them and I think we can manage running `yarn develop` :)

> Ok, the style change was applied

I've seen you merged Restyled. It's not setup properly yet. You should rely exclusively on the pre-commit hook that is installed via `yarn` and runs automatically when you do `git commit`.

> In this case a pre-rendered page displays a skeleton template while the data waits to be rehydrated with React hydration function, this deliver a super-fast experience to the users. No re-render on page transitions.

so, this how static rendering and gatsby work more or less already?

the difference is that after your changes we have to run JS code client-side now. Means that we have to pull now everything related to remark, mdx, and stuff to the client, first and means running the processing every time. Before it was a pure HTML returned from the server. The question is can we keep the same logic as before - return HTML on navigation?

sekmet:
> @shcheklein 

> I hope not. I don't like them. I don't like to support the code around them and I think we can manage running `yarn develop` :)

Ok, cool :100: 

> so, this how static rendering and gatsby work more or less already?
Yes, work more or less in this way. But still is more SSR than Pre-rendering.

- Server-side rendering allows developers to pre-populate a web page with custom user data directly on the server. It is generally faster to make all the requests within a server than making extra browser-to-server round-trips for them. This is what developers used to do before client-side rendering 

- Pre-rendering is a tradeoff between client-side and server-side rendering. Every pre-rendered page displays a skeleton template while the data waits to be rehydrated with AJAX/XHR requests. Once the page is fetched, internal routing is done dynamically to take advantage of a client-side rendered website.

> the difference is that after your changes we have to run JS code client-side now. Means that we have to pull now everything related to remark, mdx, and stuff to the client, first and means running the processing every time. Before it was a pure HTML returned from the server. 

Ok, I got. And about this:
- Will you guys keep using markdown to write the blog posts? or want to use html? 
- Into the ideal scenario you guys want remove everything related to remark and mdx processing?

> The question is can we keep the same logic as before - return HTML on navigation?
This is a good question, I think is possible :). I'm researching about  possible approach to achieve this using gatsby. As soon I have something I come back here.

Issue #44, opened 2020-01-02
determine priority and proper rendering of a hierarchy of styles


shcheklein:
E.g.

# Renders `code` with title styles

vs we do not do this:

<img width="814" alt="Screen Shot 2020-01-02 at 3 20 30 PM" src="https://user-images.githubusercontent.com/3659196/71699239-75427500-2d73-11ea-9f4d-7551525eca1d.png">

**Other examples:**

> a **bold** or _italic_ inside a quote

ksandr-dev:
Hello Ivan!, How are you today?
I have a lot experience in styling for typography in React.js. I saw you are using PostCSS. I can implement determine priority and proper rendering of a hierarchy of styles.

shcheklein:
@ksandr-dev thanks, that would be great, give it a shot!

ksandr-dev:
@shcheklein Thank you for assigning this feature to me. Are you able to discuss via skype or slack

shcheklein:
@ksandr-dev yes, find me on our Discord (dvc.org/chat) - ivan.

frontexpert:
@shcheklein I have fixed small Markdown style issues on my local. I didn't commit yet. I am not sure if this is just what you wanted or not. I may misunderstand your requirement. So I want you to take a look at this screenshot.
![screen_19](https://user-images.githubusercontent.com/19707900/73096206-f6ada300-3eec-11ea-9413-7d957b3c018c.png)

Please let me know if this is correct or not. and then, I am willing to push my result.
Thanks.

shcheklein:
@frontexpert we need to check medium and dev.to to see how do they handle different combinations - link within a title? code block within a title? link in a quote? code and link in a quote - etc.

Those examples you provided look correct! But let's check and think about other possibilities a little bit.

frontexpert:
OK, No problem. In fact, I just wanted to confirm if it is correct to have same font-style for Code and Title (In the case Title have a code part), or not.

PR #57, opened 2020-01-05
fix renders code with title styles

ksandr-dev:
@shcheklein 
#44 was fixed


shcheklein:
Closing this for now as it does not solve the problem.

Issue #61, opened 2020-01-08
card text issues


abdullahtariq1171:
### 1) text overlap
![text overlap](https://monosnap.com/image/jThEutHTShsywBhPcOkHpHaaOYr9fN)

### 2) text hidden
![ text hidden](https://monosnap.com/image/pP8DP6tqpuVXcszgUMCs0sEskLBZU3)

shcheklein:
@abdullahtariq1171 good catch! do you mind sharing what browser do you run it on? I was trying to reproduce and I can't so far.

abdullahtariq1171:
@shcheklein 
- Firefox 71.0
- Chrome 79.0.3945.88

shcheklein:
@abdullahtariq1171 but on the second screenshot text does not overlap for the same page, how is it possible?

abdullahtariq1171:
@shcheklein nah both are different issues. I put both in one issue only because they are related to card. 

Here is how we can reproduce https://monosnap.com/file/lCcyrGgj2EFil4pacIRUgNDTwgWkhK

shcheklein:
k, I see. It's not clear that width was the key to reproduce. Thanks!

fadilmamedov:
@shcheklein I will take care of this one

shcheklein:
@fadilmamedov thanks! 👍

fadilmamedov:
@shcheklein regarding the second part of the issue. I don't think that dividing the second row into two rows would look better. What about shrinking the padding of the description and meta sections in the card to 20px? Then it would look like:

<img width="723" alt="Screenshot 2020-01-09 at 21 28 17" src="https://user-images.githubusercontent.com/3175922/72094230-fb882980-3326-11ea-90df-deb75423b550.png">

In that case it first of all looks better and additionally meta information can be placed into one single row.

shcheklein:
@fadilmamedov works for me.

shcheklein:
Agreed with @cosmn to take a holistic look at how card is structured.

cosmn:
@shcheklein Latest commits and PR should close this issue.
For a feed card refactoring, we should open a new issue, and a new branch to go with that. Either way, it wont have a significant visual impact, just the a new CSS approach.

shcheklein:
Fixed by #89

PR #62, opened 2020-01-09
#61. Card text issues

fadilmamedov:


src/components/feed/item/styles.module.css
```diff
@@ -1,4 +1,6 @@
 .wrapper {
+  display: flex;
```
shcheklein: why do we need this? could you elaborate please?

shcheklein:
@fadilmamedov it looks like it breaks the main block on the large screen:

Screen Shot 2020-01-09 at 11 53 11 AM

src/components/feed/item/styles.module.css
fadilmamedov: I wanted to put meta section at the very bottom of the card, which is possible by making a card container to be displayed as `flex` and meta section with `margin-top` auto. Keeping meta section absolutely positioned causes text overlapping

fadilmamedov:
> @fadilmamedov it looks like it breaks the main block on the large screen:
> 
> Screen Shot 2020-01-09 at 11 53 11 AM

Hmmm... I can't reproduce this issue neither in Chrome nor in Firefox. In what browser were you able to reproduce this?

fadilmamedov:
> @fadilmamedov it looks like it breaks the main block on the large screen:
> 
> Screen Shot 2020-01-09 at 11 53 11 AM

Fixed! Thanks!

src/components/feed/item/styles.module.css
shcheklein: right! just curios - why was that happening on the medium screen size only?

iAdramelk:
@fadilmamedov thanks for the pull request. Couple small bugs from me:

1. In the large block if there is too much content, fade is not shown:

image

2. Second issue is still present on iPhone 5:

image

3. Margin auto didn't move author info down in Firefox (and IE11, but we can live with that one):

image

fadilmamedov:
@iAdramelk thanks a lot for your comments! I have fixed the first and the third issues you pointed out. But I am not sure about the second point. Do you think we need to take into account this issue for only iPhone 5/SE users? Because moving information about the read time will look weird. What do you think @shcheklein?

shcheklein:
@fadilmamedov moving to the next line looks weird indeed, but it's not nice that it's broken still on certain layouts. Are there any other options, ideas?

fadilmamedov:
@shcheklein ok, I have an idea. What about using a short date string, something like "Sep 05, 2019", if the row could not be placed in one single row?

shcheklein:
@fadilmamedov yep, it sounds reasonable to me. Probably it should always keep the same format though - no matter if it fits or not. It would be simpler to implement.

iAdramelk:
Thanks for the fixes! I'm ok with shorter date format.

shcheklein:
@fadilmamedov any updates on this one?

shcheklein:
closing as stale

PR #65, opened 2020-01-14
Custom html links

webdevcrazier:


content/blog/2019-03-05-march-19-dvc-heartbeat.md
```diff
@@ -36,18 +36,13 @@ checking.
...
-    </div>
-  </section>
-</a>
+<external-link
+href="https://veekaybee.github.io/2019/02/13/data-science-is-different/"
+title='Data science is different now'
+description="Woman holding a balance, Vermeer 1664 What do you think of when you read the phrase 'data science'? It's probably some…"
+link='veekaybee.github.io'
+image='/uploads/images/2019-03-05/data-science-is-different-now.png'>
+</external-link>
```
shcheklein: can we avoid putting the closing html tag `</external-link>` everywhere?

gatsby-config.js
```diff
@@ -69,7 +70,8 @@ const plugins = [
           }
         },
         'gatsby-remark-copy-linked-files',
-        'gatsby-remark-smartypants'
+        'gatsby-remark-smartypants',
+        'external-linc-plugin'
```
shcheklein: linc -> link?

plugins/external-linc-plugin/index.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+  return a;
+}
+
+module.exports = ({ markdownAST }) => {
+  visit(markdownAST, 'html', node => {
+    if (node.value.includes('<external-link')) {
+      const found = node.value.match(re);
+      const attrs = found.reduce((res, item) => {
+        const objectParse = item.split('=');
```
shcheklein: can `=` be inside the value?

plugins/external-linc-plugin/index.js
```diff
@@ -0,0 +1,47 @@
+const visit = require('unist-util-visit');
+
+const re = /((\S+)=["']?((?:.(?!["']?\s+(?:\S+)=|[>"']))+.)["']?)/g;
```
shcheklein: it looks too complicated. Put a comment what it does

content/blog/2019-03-05-march-19-dvc-heartbeat.md
```diff
@@ -36,18 +36,13 @@ checking.
...
-      <div class="elp-link">veekaybee.github.io</div>
-    </div>
-    <div class="elp-image-holder">
-      <img src="/uploads/images/2019-03-05/data-science-is-different-now.png" />
-    </div>
-  </section>
-</a>
+<external-link
+href="https://veekaybee.github.io/2019/02/13/data-science-is-different/"
+title='Data science is different now'
```
iAdramelk: Small nitpick. Can we use `"` everywhere in the `.md` files? Right now there is a mix between `'` and `"` in different attributes.

plugins/external-linc-plugin/index.js
```diff
@@ -0,0 +1,47 @@
...
+      const attrs = found.reduce((res, item) => {
+        const objectParse = item.split('=');
+        return {
+          ...res,
+          [objectParse[0]]: stripquotes(objectParse[1])
+        };
+      }, {});
+
+      node.type = 'html';
+      node.value = `
```
iAdramelk: This will work if `<extenal-link>` is the only tag is the html node. But it can potentially be a child of other node and have sibling nodes, e. g.:

 ```
<div class="some-class">
    <external-link />
    <div>Some text</div>
</div>
```

Replacing `node.value` will delete wrapper and siblings.

I think we should use `rehype` here to find `external-link` node and replace only it. 

We should probably also validate it's fields here. Is some of the required attributes are missing, we should throw error and stop build.

File: plugins/external-linc-plugin/index.js [hunk 7]
```diff
@@ -0,0 +1,47 @@
...
+        <div>
```
iAdramelk: I think div wrapper can be removed here.

plugins/external-linc-plugin/index.js
```diff
@@ -0,0 +1,47 @@
...
+          <a href=${attrs.href} class="external-link-preview">
+            <section class="elp-content-holder">
+              <div class="elp-description-holder">
+                <h4 class="elp-title">${attrs.title}</h4>
+                <div class="elp-description">${attrs.description}</div>
+                <div class="elp-link">${attrs.link}</div>
+              </div>
+              <div class="elp-image-holder">
```
iAdramelk: I think that `image` is an optional field and can be omitted sometimes. So if there is no `image` field, this block shouldn't be added.

plugins/external-linc-plugin/index.js
```diff
@@ -0,0 +1,47 @@
...
+                <img src="${attrs.image}"/>
```
iAdramelk: Images should always have an `alt` field, even if it is a blank one.

plugins/external-linc-plugin/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "external-linc-plugin",
+  "version": "1.0.0",
+  "main": "index.ts",
+  "author": "",
+  "license": "MIT",
+  "dependencies": {
+    "unist-util-visit": "^2.0.1"
```
iAdramelk: I'm not totally sure here, but can't local plugins work with global `package.json` file? Or do you want to release this plugin as an separate library?

content/blog/2019-03-05-march-19-dvc-heartbeat.md
webdevcrazier: Yes, we can.

gatsby-config.js
webdevcrazier: fixed

plugins/external-linc-plugin/index.js
webdevcrazier: fixed

plugins/external-linc-plugin/index.js
webdevcrazier: this is an expression for finding the key and value of attributes in external-link, for example, it finds attr link='martinfowler.com' in <external-link link='martinfowler.com' />

content/blog/2019-03-05-march-19-dvc-heartbeat.md
webdevcrazier: sure

plugins/external-linc-plugin/index.js
webdevcrazier: fixed

plugins/external-linc-plugin/index.js
webdevcrazier: this div for correct render

plugins/external-linc-plugin/index.js
webdevcrazier: fixed

plugins/external-linc-plugin/index.js
webdevcrazier: fixed

plugins/external-linc-plugin/package.json
webdevcrazier: fixed

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+  return value;
+}
+
+function isExternalLinkAttr(elem) {
+  return attrKeyArray.some(item => elem.includes(item));
+}
+
+function renderTag(withImage, attrs) {
+  if (withImage) {
```
shcheklein: need to rewrite this to avoid duplication

plugins/external-link-plugin/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "external-link-plugin",
+  "version": "1.0.0",
+  "main": "index.ts",
+  "author": "",
+  "license": "MIT"
```
shcheklein: license is Apache 2 for the project

content/blog/2019-03-05-march-19-dvc-heartbeat.md
shcheklein: 👍

plugins/external-linc-plugin/package.json
shcheklein: how is it fixed now?

webdevcrazier: deleted dependency from plugin's package.json and deleted yarn.lock file. Global project node modules are used in this case

shcheklein: I think the question was if we can avoid creating a separate "package.json" in the first place? If we have to create I would keep dependencies in it as it was before - it's better to have this level of isolation.

plugins/external-link-plugin/package.json
Eric32112: fixed it

webdevcrazier:
@shcheklein what did you mean at all about dependencies in plugin?

plugins/external-linc-plugin/package.json
iAdramelk: Yes, I thought that we can remove `package.json` altogether and rename plugin as `external-link-plugin.js`, but I read manual one more time and looks like it's not the case. We still need `package.json`. If we still need it, then I'm also ok with leaving unist there. Sorry about that.

plugins/external-linc-plugin/index.js
iAdramelk: Can you tell in more detail what exactly is it doing? There is no div in original markdown and if I understand correctly we are here just insert html string. Was `<p>` was added by parser automatically in the past because of new lines?

iAdramelk: Can we achieve the same effect by adding styles to the `<section>` or `<a>` instead?

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,68 @@
...
+        </section>
+      </a>
+    </div>`;
+}
+
+module.exports = ({ markdownAST }) => {
+  visit(markdownAST, 'html', node => {
+    if (node.value.includes('<external-link')) {
+      const found = node.value.match(re);
+      if (found.every(isExternalLinkAttr)) {
```
iAdramelk: This will check if all attributes are correct attributes, but it will not check if we have all attributes. It will not throw error if `href` is missing for example.

plugins/external-linc-plugin/index.js
iAdramelk: I think the problem of nested tags is still not fixed. (wrapper tag and siblings are still deleted).

Validation now checks that only allowed attributes are exists, but it didn't throw error if some of required attributes are missing – for example it should throw error if we don't have `href` or `title`.

plugins/external-linc-plugin/index.js
webdevcrazier: @iAdramelk moved the <section> up <a> and changed styles

plugins/external-linc-plugin/package.json
webdevcrazier: done

plugins/external-linc-plugin/index.js
webdevcrazier: done

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,66 @@
...
+  each(collection);
+}
+
+module.exports = ({ markdownAST }) => {
+  visit(markdownAST, 'html', node => {
+    const data = unified()
+      .use(parse)
+      .parse(node.value);
+    getExternalLinkTag(data.children, ['children'], value => {
+      if (_.includes(['external-link'], value.tagName)) {
```
iAdramelk: Why do we need lodash here? Can't we just write it as `value.tagName === 'external-link'`?

iAdramelk:
@webdevcrazier thanks for the updates! I added one question, but everything else looks good to me.

plugins/external-link-plugin/index.js
webdevcrazier: lodash deleted, I used hast-util-select

plugins/external-link-plugin/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "external-link-plugin",
+  "version": "1.0.0",
+  "main": "index.ts",
+  "author": "",
+  "license": "Apache 2",
```
shcheklein: please, change it to `Apache-2.0`

package.json
```diff
@@ -43,11 +43,13 @@
     "gatsby-plugin-twitter": "^2.1.17",
     "gatsby-remark-embed-gist": "^1.1.9",
     "gatsby-remark-relative-images": "^0.2.3",
+    "hast-util-select": "^3.0.1",
```
shcheklein: are they needed only for the plugin? move them to the `plugins/external-link-plugin/package.json`

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,55 @@
...
+    </section>
+    `;
+}
+
+module.exports = ({ markdownAST }) => {
+  visit(markdownAST, 'html', node => {
+    const data = unified()
+      .use(parse)
+      .parse(node.value);
+      const externalLink = select('external-link', data);
```
shcheklein: what happens if there are a few of those?

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,55 @@
...
+      if(externalLink) {
+        const { properties } = externalLink;
+        if (isCorrectExternalLinkAttr(Object.keys(properties))) {
+          let isImage = Boolean(properties.image);
```
shcheklein: let -> const?

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,55 @@
+const visit = require('unist-util-visit');
+const unified = require('unified');
+const parse = require('rehype-parse');
+const select = require('hast-util-select').select;
+
+const attrKeyArray = ['href', 'title', 'description', 'link'];
+
+function isCorrectExternalLinkAttr(elem) {
```
shcheklein: is `elem` an array? name is confuising

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    attrKeyArray.filter(item => elem.some(value => value.includes(item)))
```
shcheklein: there should be a way simpler way to test this

also, `value.includes(item)` - what is value here and what is item?

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,55 @@
...
+    const data = unified()
+      .use(parse)
+      .parse(node.value);
+      const externalLink = select('external-link', data);
+      if(externalLink) {
+        const { properties } = externalLink;
+        if (isCorrectExternalLinkAttr(Object.keys(properties))) {
+          let isImage = Boolean(properties.image);
+          node.type = 'html';
+          node.value = renderTag(isImage, properties);
```
shcheklein: it means that we are replacing it completely, right? if it's nested inside some other HTML it will break

webdevcrazier: we are replacing only <external-link /> tag

plugins/external-link-plugin/index.js
webdevcrazier: fixed

plugins/external-link-plugin/index.js
webdevcrazier: fixed

plugins/external-link-plugin/index.js
webdevcrazier: fixed

package.json
webdevcrazier: there were problems with pull-request checks, during auto deploy has error 'cannot find this module'
Screenshot from 2020-01-20 22-11-00

plugins/external-link-plugin/package.json
webdevcrazier: fixed

plugins/external-link-plugin/index.js
webdevcrazier: there are only one tag external-link from data

package.json
shcheklein: so, why is it happening? we need to figure out this.

webdevcrazier: @shcheklein fixed this problem

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,56 @@
+const visit = require('unist-util-visit');
```
shcheklein: @iAdramelk can we make this file type-script?
@iAdramelk is it a regular way to import dependencies? (with require and `const smt =`?)

package.json
```diff
@@ -12,7 +12,8 @@
     "typecheck": "tsc --noEmit",
     "lint-ts": "tslint --project .",
     "lint-css": "stylelint \"src/**/*.css\"",
-    "format-all": "prettier --write './**/*.{js,jsx,md,tsx,ts,json}'"
+    "format-all": "prettier --write './**/*.{js,jsx,md,tsx,ts,json}'",
+    "postinstall": "cd plugins/external-link-plugin/ && yarn install"
```
shcheklein: I don't get it. Why can't we specify a dependency in the dependencies section so it installs when you just do `yarn` ... can you link to some doc about this?

webdevcrazier: https://spectrum.chat/gatsby-js/general/local-plugin-dependencies-not-installed-during-build~361cf7a6-0723-4f74-85f4-f1f4e3ba31da

shcheklein: what happens in the dev mode? if I change the code of the plugin will I have to reinstall again?

shcheklein: also what about this - https://spectrum.chat/gatsby-js/general/local-plugin-dependencies-not-installed-during-build~361cf7a6-0723-4f74-85f4-f1f4e3ba31da?m=MTU2NDgzMTg1NjA2OA== ?

plugins/external-link-plugin/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "external-link-plugin",
+  "version": "1.0.0",
+  "main": "index.ts",
```
shcheklein: index.js?

plugins/external-link-plugin/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "author": "",
```
shcheklein: put your name

shcheklein: or actual author name )

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,55 @@
...
+    const data = unified()
+      .use(parse)
+      .parse(node.value);
+    const externalLink = select('external-link', data);
+    if (externalLink) {
+      const { properties } = externalLink;
+      if (isCorrectExternalLinkAttr(Object.keys(properties))) {
+        const isImage = Boolean(properties.image);
+        node.type = 'html';
+        node.value = renderTag(isImage, properties);
```
shcheklein: as we discussed there is still a problem with nested nodes, and probably with multiple nodes as well

plugins/external-link-plugin/index.js
iAdramelk: @shcheklein it will be ok this time. This is a separate html ast that we created from html string. We are replacing external-link tag in it.

plugins/external-link-plugin/index.js
iAdramelk: Yep,  this file will be used by nodejs without transpiling so we need to write it in the way that nodejs will understand. We can potentially add typascript compilation here, but it will mostly overcomplicate stuff.

package.json
iAdramelk: Let's just move dependencies to main package.json.

plugins/external-link-plugin/index.js
iAdramelk: My bad, read code wrong. @shcheklein is right, we need to replace `external-link` in rehype ast, stringify it and the upend stringified result to `node.value`.

plugins/external-link-plugin/index.js
webdevcrazier: @iAdramelk 
So do you want to add typescript or not?

package.json
shcheklein: Agreed, if there are no way around, let's at least keep build scripts simple. Would be great if we can specify a dependency "group" in the main package json.

plugins/external-link-plugin/index.js
shcheklein: Resolving this. Thanks @iAdramelk

plugins/external-link-plugin/package.json
webdevcrazier: added author

plugins/external-link-plugin/package.json
webdevcrazier: fixed it

package.json
webdevcrazier: moved to main package.json

shcheklein: so, remove this postinstall then?

plugins/external-link-plugin/yarn.lock
```diff
@@ -0,0 +1,198 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
shcheklein: remove this file, merge master, update the main one yarn lock

plugins/external-link-plugin/index.js
iAdramelk: @webdevcrazier Nope, we should leave it as is.

plugins/external-link-plugin/yarn.lock
webdevcrazier: done

package.json
webdevcrazier: done

plugins/external-link-plugin/index.js
webdevcrazier: if some html includes external-link, we should replace only <external-link/> node and left outer html without changes, right ?

shcheklein: ideally yes, as a plan minimum detect and throw an exception in those situations

webdevcrazier: i've updated logic of external-link-plugin.
In short I convert HTML in md to HAST (Hypertext Abstract Syntax Tree), then find all "external-link" in HAST, convert them to equivalent HAST of external-link html, then convert updated HAST back to html string and that's it

webdevcrazier: important note:
if we put external-link inside other html, we shouldn't write self-closed tag like this <external-link/>. "rehype-parse" library has issue with self-closed tags. We should write <external-link ...></external-link> - with empty body

.gitignore
```diff
@@ -4,4 +4,5 @@ public
 yarn-error.log
 tmp/*
 .env
-.DS_Store
\ No newline at end of file
+.DS_Store
+.idea
```
shcheklein: please, add a new line at the end

package.json
shcheklein: thanks!

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,74 @@
...
+    externalLinkNodeList.forEach(externalLinkNode => {
+      const { properties } = externalLinkNode;
+      if (isCorrectExternalLinkAttr(Object.keys(properties))) {
+        const externalLinkHtml = renderTag(properties);
+        const externalLinkHast = convertHtmlToHast(externalLinkHtml);
+
+        externalLinkNode.type = externalLinkHast.type;
+        externalLinkNode.tagName = externalLinkHast.tagName;
+        externalLinkNode.properties = externalLinkHast.properties;
+        externalLinkNode.children = externalLinkHast.children;
```
shcheklein: @iAdramelk could you take a look please? it looks right, just want to make sure that we don't break anything by hijacking directly the `externalLinkNode` object. Is it better to find parent and replace the node with `externalLinkHast`?

plugins/external-link-plugin/index.js
```diff
@@ -0,0 +1,74 @@
...
+      } else {
+        throw new Error(
+          `No correct tag <external-link /> or not all nested tags in ${node.value}`
+        );
+      }
+    });
+
+    node.value = convertHastToHtml(hast);
```
shcheklein: same here ... is it a regular/normal protocol to replace value directly (looks unsafe) in the visitor? cc @iAdramelk ?

.gitignore
webdevcrazier: Fixed

plugins/external-link-plugin/index.js
iAdramelk: @shcheklein yep, they use it in tutorial, so I think we can call it recommended method.

plugins/external-link-plugin/index.js
iAdramelk: Yep, see above.

PR #66, opened 2020-01-15
Fix: create a custom markup for inline blocks

oleksandrbyk:
@shcheklein 
#16 is done on this PR.
It is done all md transformation on gatsby-node.js and get the fully transformed HTML content from graphql query as original HTML content would be.


shcheklein:
Closing in favor of https://github.com/iterative/blog/pull/65

PR #69, opened 2020-01-17
Changed post image size

webdevcrazier:


src/templates/blog-post.tsx
```diff
@@ -107,7 +107,7 @@ export const pageQuery = graphql`
         }
         picture {
           childImageSharp {
-            fluid(maxWidth: 1000, maxHeight: 750, cropFocus: CENTER) {
+            fluid(maxWidth: 1500, cropFocus: CENTER, quality: 80) {
```
shcheklein: why is it 1500?

webdevcrazier: For better resolution, I can implement (original width image / 2 ) but what about when the image is too large or too small, so I wanted to be consistent. If you prefer ( original / 2 ). I will make it.

shcheklein: so, what does exactly `maxWidth: 1500` mean here?

webdevcrazier: It means gatsby generates picture with that width if it's less than that It will just take the original width so we don't get ugly pixel like resolution it optimizes them.

shcheklein: so, why do we want 1500 if in other places we reduced the width to < 1000?

webdevcrazier: For better picture resolution in blog, if you think it's unnecessary I will return the 1000.But in my opinion it's better like that so images have bigger pixel resolution.

shcheklein: again, I don't understand what does this option exactly do I think. Could you explain the whole image workflow - from the original one that has certain original dimension in pixels to the end result? And where does this change fit into?

webdevcrazier: Create flexible size for an image that stretches to fill its container. E.g. for a container whose max width is 800px, this change creates the images with 1500 width so they can be as you said original / 2 that is why.
For example we have a image with a width of 2500 it's unnecessary to put that kind of big image in a section with 800px it will take more time to load and make page heavier. So it decrease the image, however you wanted original image size / 2 so I increased the width to get better looking images.

shcheklein: I don't think it works like this. In max-width you need to specify the maximum container width. So, if you decided to reduce it, you should reduce the max width here as well.

iAdramelk: Gatsby-image automatically create @x2 images out of the box. See https://blog.dvc.org main image. Its maxWidth is set to 650px, but on retina display downloaded image width is 1300px.

webdevcrazier: Okay. Fixed it changed it to the max width which container takes.

webdevcrazier: @iAdramelk 
Do you have any idea of more improvement ?

iAdramelk: Missed the moment then image became 850px instead of 1000px. But yes, if we reduced image container, we should also reduce image maxWidth. Not sure about `quality: 80` through, default 50 should be ok for us.

webdevcrazier: @iAdramelk 
I made new PR regarding your suggestion
https://github.com/iterative/blog/pull/79

Issue #39, opened 2019-12-31
support basic markup in meta fields


shcheklein:
For example, we need to put links into `pictureComment` or support newlines/paragraph splits in `descriptionLong`.

shcheklein:
Example of links in the title image: https://blog.dataversioncontrol.com/june-19-dvc-%EF%B8%8Fheartbeat-505900be9aac

shcheklein:
Example of newlines in the "description long" - https://blog.dataversioncontrol.com/june-19-dvc-%EF%B8%8Fheartbeat-505900be9aac

iAdramelk:
@shcheklein It's technically possible, but we will have additional problems then. For example `descriptionLong` is used for the meta tags and you can't have tags there. And if we remove tags from there automatically it can came out awkward: Texts instead of links, multiple paragraphs that became one paragraph, etc. What is the best way to approach that?

sekmet:
@shcheklein I got fixed #25 and #39

shcheklein:
@iAdramelk can we strip markup when we need to put those into meta fields?

iAdramelk:
@shcheklein we can, but text can loose meaning if we replace link with text or concat paragraphs.

shcheklein:
@iAdramelk I think links to text and concatenation should work fine? why do you have concerns about this? any specific examples in your mind?

iAdramelk:
@shcheklein the are no technical problems, only content one.

For example if we write something like that in description:

> Please check [this link].
>
> Some new paragraph.

And then replace it with:

> Please check this link. Some new paragraph.

And then place it in say feed or in the Facebook snippet it can look strange and meaningless.

shcheklein:
yep, agreed. But still does not look like a deal breaker to me. Newlines are fine. Link - hard to predict but should be more or less fine either (I doubt that we will have [this link] type of phrase in that block - it's too important). Links in image subtitles - probably fine as well.

fadilmamedov:
@shcheklein please, assign this to me. Thanks!

PR #72, opened 2020-01-20
#39. Support basic markup in meta fields

fadilmamedov:


src/components/post/index.tsx
```diff
@@ -1,5 +1,6 @@
 import cn from 'classnames';
 import Image from 'gatsby-image';
+import ReactMarkdown, { NodeType as MarkdownNodeType } from 'react-markdown';
```
shcheklein: @fadilmamedov it means that we bring react markdown and everything to the client side just to render these meta fields, right? This is pretty bad. We should do the pre-processing on the rendering side. Also, check the comments from Alex in the ticket, we should flatten/render these field when we populate meta for the page.

shcheklein:
closing this due to inactivity and approach itself is not correct here

PR #73, opened 2020-01-21
January '20 heartbeat

andronovhopf:


content/blog/2020-01-17-january-20-dvc-heartbeat.md
```diff
@@ -0,0 +1,164 @@
+---
+title: January '20 DVC❤️Heartbeat
+date: 2020-01-17
+description: |
+  Reaching 100 contributors, PyData LA,
```
shcheklein: minor: we try to wrap at 80 symbols. Probably no need to be aggressive here.

PR #85, opened 2020-01-27
improve highlighting

### LINKED ISSUE #82: syntax highlighting showing bugs
DVC syntax highlighting in code blocks seems to have some errors- 

Screen Shot 2020-01-24 at 1 19 24 PM

fabiosantoscode:
I've added a bunch of syntax to the prism language. To keep the user from selecting the `$ ` part of commands I added a prism hook, which acts after tokenisation and changes the token stream.

The for loop support isn't stellar, but since I don't see us using too many for loops I didn't spend too much time on this. However it's possible to improve the support if we are OK with not highlighting the "do" keyword, which I think is quite a bit of a loss.

Fixes #82 

## Before

Screenshot from 2020-01-27 19-18-59

## After

Screenshot from 2020-01-27 19-18-05

### DIFF
--- a/config/prismjs/dvc.js
+++ b/config/prismjs/dvc.js
@@ -1,3 +1,11 @@
+// we require prism and load its bash module so we have
+// Prism.languages.bash to embed into our DVC language.
+
+const Prism = require('prismjs');
+require('prismjs/components/prism-bash');
+require('./dvc-hook');
+const { bash } = Prism.languages;
+
 // Command arrays are intentionally reverse sorted
 // to prevent shorter matches before longer ones
 
@@ -71,24 +79,31 @@ const dvc = [
 
 const beforeCommand = String.raw`(\$[\s(]+|;\s*)`;
 
-module.exports = {
-  dvc: {
-    pattern: new RegExp(
-      String.raw`${beforeCommand}\b(?:dvc (?:${dvc.join('|')}))\b`
-    ),
-    greedy: true,
-    lookbehind: true
+Prism.languages.dvc = {
+  line: {
+    pattern: /(?<=(^|\n))\$[\s\S]*?[^\\](:?\n|$)/,
+    inside: {
+      dvc: {
+        pattern: new RegExp(
+          String.raw`${beforeCommand}\b(?:dvc (?:${dvc.join('|')}))\b`
+        ),
+        greedy: true,
+        lookbehind: true
+      },
+      git: {
+        pattern: new RegExp(
+          String.raw`${beforeCommand}\b(?:git (?:${git.join('|')}))\b`
+        ),
+        greedy: true,
+        lookbehind: true
+      },
+      command: {
+        pattern: new RegExp(String.raw`${beforeCommand}\b[a-zA-Z0-9\-_]+\b`),
+        greedy: true,
+        lookbehind: true
+      },
+      ...bash
+    }
   },
-  git: {
-    pattern: new RegExp(
-      String.raw`${beforeCommand}\b(?:git (?:${git.join('|')}))\b`
-    ),
-    greedy: true,
-    lookbehind: true
-  },
-  command: {
-    pattern: new RegExp(String.raw`${beforeCommand}\b[a-zA-Z0-9\-_]+\b`),
-    greedy: true,
-    lookbehind: true
-  }
+  comment: bash.comment
 };
--- a/config/prismjs/dvc.js
+++ b/config/prismjs/dvc.js
@@ -71,17 +71,7 @@ const dvc = [
 
 const beforeCommand = String.raw`(\$[\s(]+|;\s*)`;
 
-const variable = {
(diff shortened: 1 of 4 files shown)


gatsby-config.js
```diff
@@ -1,5 +1,20 @@
 const path = require('path');
 const dvc = require('./config/prismjs/dvc');
+const Prism = require('prismjs');
+
+// Make sure the $ part of the command prompt in shell
+// examples isn't copiable by making it an 'input' token.
+Prism.hooks.add('after-tokenize', env => {
```
shcheklein: can we the code to a separate file? just want to keep the config clean more or less

fabiosantoscode: Sure!

shcheklein:
@fabiosantoscode  minor: color scheme is kinda strange. let's do dvc command colored as git now, and git with some other color, similar to how bash command are highlighted now on dvc.org

shcheklein:
probably it's fine to highlight `pythion` as a command - we do the same for dvc.org for example. If it's a different color with dvc and git commands

fabiosantoscode:
I admit I'm terrible at picking colors :) I'll use those.

WRT python, python is highlighted. Just not `dvc add python` and such.

fabiosantoscode:
@shcheklein I hope this is what you meant. I made it highlight anything that looks like a command at the start of a line (`[a-zA-Z0-9]+`). `git` and `dvc` are special though and have different colors from each other. `dvc` has the same color `git` had, and I made `git` have a different color, this time more related to the regular command color.

Screenshot from 2020-01-28 11-34-46

fabiosantoscode:
After pushing that last commit I devised a color scheme for the commands where they have more similar brightness and saturation. Let me know if you find this one better.

Screenshot from 2020-01-28 11-42-54

fabiosantoscode:
Maybe "git" and "dvc" commands should be in similar shades of blue (or the same one) to identify they're both for version control and tell them apart from other commands.

config/prismjs/dvc.js
```diff
@@ -1,37 +1,14 @@
 // Command arrays are intentionally reverse sorted
 // to prevent shorter matches before longer ones
 
-const commands = [
```
shcheklein: why do we remove this?

fabiosantoscode: Because now what defines a command is:

1. It's at the start of the input (after `$ `, `$(` or `;`)
2. It looks like a command

This way any new command we might start to use in the docs should work without modifying this list.

shcheklein: okay, that's fine ... ideally we should probably take some `bash` highlighter, use it for each line and modify a bit to detect dvc and git commands so that we hight

fabiosantoscode: prismjs comes with a bash highlighter, and it has ways to extend existing languages. It wasn't an option for me before because of the "input" token but now that I know I can massage the token stream it could be a valid approach.

shcheklein: do you want to take a look and give it a try? I just wonder if we can do something more holistic there.

fabiosantoscode: I'll have a look after dinner :) 

BTW, have a look at the colors in here and see what you think: https://github.com/iterative/blog/pull/85#issuecomment-579206815

I'm not good at picking colors so I don't want to decide on this one.

shcheklein: yep, they are absolutely fine. And we can adjust them a bit when it's needed.

shcheklein:
@fabiosantoscode this looks great! a bit of complication though :) We should be taking into account somehow lines that are outputs. They are not bash. We need to apply a slightly different style to them.

fabiosantoscode:
@shcheklein that's a shame, it was pretty simple. I had to change my approach a little bit. Instead of extending bash, I reverted back to our old syntax (with "line") and embedded prism's own bash definition into the line (after our special highlighting on commands). Additionally I enabled comments outside of the line, so that you can write something like:

```
# this command does X
$ do-x
```

And have the command properly highlighted.

shcheklein:
@fabiosantoscode looks simpler, that's good :) check the existing styles (especially dvc.org) - we need to use a different font weight for outputs (may be comments as well?).

(and it should understand that where is it output where it's not)

fabiosantoscode:
@shcheklein adjusted the colors and font-weights. It would be hard to distinguish comments from output with font-weight alone, so I changed the color a bit.

Had to use `!important` to get some specificity in order without restructuring a lot of stuff. I won't sleep well tonight.

shcheklein:
@fabiosantoscode looks perfect to me! not sure why did you do `important!` though - any way around it?

fabiosantoscode:
@shcheklein this was a bit dumb, sorry.

Moved the rule out of `:global` and set language-dvc to have the same specificity as every `language-*`.

shcheklein:
thanks @fabiosantoscode !

Issue #70, opened 2020-01-18
setting image size in markdown


andronovhopf:
I'm not able to change the size of images using markdown. For example:

```
![](/uploads/images/img.png =250x250)
```
seems to throw an error and no image displays (only a little icon). To my knowledge, this is the markdown syntax for resizing images?

fadilmamedov:
@shcheklein can you please assign this to me?

frontexpert:
@shcheklein I looked to solution for issue but now the plugin that this project is using has not this option of resizing the image into markdown! 

How we can see https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-images/src/__tests__/index.js

My proposal to a solution to this issue is:

- 1º Switch  `![](/image.ext)` to `<img src=''/image.ext' class='custom-class-dvc' width='200' height='200'>` and create a custom calss css to adjust
> but this is no better solution because we lose the beneficite of gatsby-image and gatsby-shape plugins

How we can see here in the screenshot
![Screenshot_2020-01-28_07-46-37](https://user-images.githubusercontent.com/19707900/73257458-921b6e00-41a2-11ea-8a48-3df44ad7c06d.png)
 ---

- 2º Switch transformer MD to MDX in this way we get the better of between about markdown and React
We'll can create a component React to the custom image 

```markdown
![test image size](/uploads/images/2017-05-15/git-server-or-github.jpeg)
```

```JSX
<ImageDvc width='200' height='200' position='left' > 
```

> but this solution is very hard because this is can break anything in this blog  and we'll need with this

What do you think about this? Do you have another solution to this issue? tell me

shcheklein:
Don't like both options, tbh.

Can we write our own custom plugin for this? The same way we do for the "external-link" now, for example.

frontexpert:
OK, I will try to find out the way just like "external-link". Thanks.

frontexpert:
@shcheklein I did custom the gatsby-remark-images plugin.
Please take a look at my result.
![Screenshot at Jan 30 12-25-36](https://user-images.githubusercontent.com/19707900/73441541-d3a34900-435b-11ea-9f2e-aedf6792887e.png)

You can append `=WIDTHxHEIGHT` after Title Text to resize the image.
ex:
`![](./pic/pic100.png "Img Title =100x200")`
You can skip the HEIGHT
`![](./pic/pic100.png "Img Title =100")`

shcheklein:
fixed by #92

PR #88, opened 2020-01-30
Feat/markdown image size 70

frontexpert:
This is solution for the issue #70 
I've customized the gatsby-remark-images plugin to resize images in markdown.

You can append "=`WIDTH`x`HEIGHT`" after image title text.
ex:
``
You can skip the `HEIGHT`
``


frontexpert:
@shcheklein Please check my result and let me know your feedback.
Thanks

frontexpert:
@shcheklein I just got your message.
> looks like somewhat an overkill to bring the whole plugin to us

Then, I am willing to upload my custom plugin to npm in public.
After that, I can update package.json file. so we can install the customized plugin(remark-images) instead of the original *gatsby-remark-images* plugin.

What about this?
Is this still not good solution?

Actually, Gatsby don't have any plugin for this solution now.
https://github.com/gatsbyjs/gatsby/issues/2609
This is why I customized the gatsby-remark-images now.

shcheklein:
@frontexpert the problem with this approach is that we start depending on a third party plugin for a some core functionality. We'll need to support it on our own, keep it updated, etc. That's why this an overkill for this task.

frontexpert:
OK, I got it. Thanks for your inform. Let me try again.

frontexpert:
@shcheklein 
Please check my result again.

shcheklein:
@frontexpert no, it's still not acceptable as a solution. It's still an overkill here. And we don't want to replace a regular plugin with our own version here.

If you don't see any other options in terms of implementing this custom markup, let's try to make it avoid stretching images (if it's 400x400 it should show it as 200x200 in browser, no stretching should happen). It might be enough to start. This way we will be able at least to pick the original size of the image.

frontexpert:
@shcheklein I did not find another solution to implementing this custom markup. 

>  let's try to make it avoid stretching images (if it's 400x400 it should show it as 200x200 in browser, no stretching should happen). It might be enough to start. This way we will be able at least to pick the original size of the image.

Can you explain it more, please

shcheklein:
@frontexpert custom plugin complexity should be reasonable. `external link` plugs nicely, it does not replace anything and it's pretty small and simple in terms of implementation. Similar approach if it works is acceptable. Replacing the whole `image-gatsby` is probably not.

PR #89, opened 2020-01-30
Fix: card text issues #61 Part 2

cosmn:
Issue #61 Part 2
Card inner padding decreased from 30px to 20px.
Avatar size from 30px to 40px to match 2 lines at 20px line-height each.
Date format is changed to reduce the number of characters on the 2nd line.
Post meta area padding is adjusted as well to fit the new spacing

Avatar changes are reflected in the single post page as well, without messy side effects.

### DIFF
--- a/src/components/feed/item/styles.module.css
+++ b/src/components/feed/item/styles.module.css
@@ -86,7 +86,7 @@
 .body {
   position: relative;
   overflow: hidden;
-  padding: 10px 30px;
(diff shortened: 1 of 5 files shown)


shcheklein:
Good stuff! Merging and releasing this. Created #90 to fix avatars - they are blurry on the retina display, also need to add CENTER to the sharp image processor to make sure that it takes center.

Issue #90, opened 2020-01-30
fix avatars quality


shcheklein:
Looks like some images are too low res and/or something wrong with the gatsby pipeline here

shcheklein:
closed by https://github.com/iterative/blog/pull/91 (@cosmn please add Fixes #90 to the next PR description to connect and close automatically).

PR #91, opened 2020-01-31
Fix: avatar missing quality option

cosmn:
Although the sharp library has a [default jpeg quality of 80](https://sharp.pixelplumbing.com/api-output#jpeg), the [gatsby-plugin-sharp](https://www.gatsbyjs.org/packages/gatsby-plugin-sharp/) ( used internally by [gatsby-image](https://www.gatsbyjs.org/packages/gatsby-image/) ) for some reason [defaults it to 50](https://www.gatsbyjs.org/packages/gatsby-plugin-sharp/#parameters), if the value is omitted.

I decided to go with a 95 quality value which seemed most adequate based on this visual test I prepared. Opinions are welcomed.
tests

Side notes:
1. Gatsby [generates new images](https://www.gatsbyjs.org/packages/gatsby-image/#some-other-stuff-to-be-aware-of) at new locations when parameters are changed.
This [may impact SEO](https://github.com/gatsbyjs/gatsby/issues/13742) and / or cause other disruptions.
2. Iphone 6+ and above have `@3x` retina display. We currently provide `@1x @1.5x @2x` for the srcset. Opinions on supporting `@3x` ? Is it worth it ?

Issue #90


cosmn:
> @shcheklein on #89 
> they are blurry on the retina display, also need to add CENTER to the sharp image processor to make sure that it takes center.

I've added the `cropFocus: CENTER`, and tested it locally with a wide image and a tall image. Works as expected

src/pages/index.tsx
```diff
@@ -116,7 +116,12 @@ export const pageQuery = graphql`
                   name
                   avatar {
                     childImageSharp {
-                      fixed(width: 40, height: 40) {
+                      fixed(
+                        width: 40
+                        height: 40
+                        quality: 95
```
shcheklein: probably not need to change quality here, default usually works fine.

shcheklein:
the problem with avatar images is visible in the older posts, please check those. 2x is fine, but it looks like we don't have large enough (80px+) images in the first place - I would check that first.

src/pages/index.tsx
cosmn: the default value is 50 as described [here](https://www.gatsbyjs.org/packages/gatsby-plugin-sharp/#parameters) which is quite low, 95 seems like the best option in my opinion

shcheklein: kk, it's fine for avatars - they won' become too large (I would compare sizes though) .. it might we don't gain anything from 80+

cosmn: Honestly there's no significant visual difference between quality 50 and 80 at 40x40 px. I guess we'll go with 50.

shcheklein:
@cosmn one more image left I belive `Marija Ilić`. Otherwise looks great, thanks for taking care of this.

iAdramelk:
Usually, for retina images we don't need high quality jpeg images, see: https://pieroxy.net/blog/2016/05/01/jpeg_compression_is_80_the_magic_quality_part_1_the_retina_screens.html

Basically, in a lot of cases quality of 30-40 and 80 are indistinguishable on retina dispalys for the end users. 50 is already having some additional buffer and, because we can't set quality for non-retina displays separately, is a fallback quality for non-retina display to make them look ok.

The problem in original ticket was with original avatar quality, not with the gatsby options, so we can leave them as default. and only change cropFocus.

cosmn:
@shcheklein Thanks, I missed that, just made a4d3cd1b48da9ed9012277bf2cdf0fd5ec5f48c4

@iAdramelk we have straighten it out. The issue was misunderstood from the beginning, hence the wrong headed direction I took.

The PR can be closed and merged now.

Issue #40, opened 2020-01-01
properly size post images


shcheklein:
Effective image size should 1/2 to make it sharp on retina screens. I suggest we always (?) make size 1/2 before we cut it into a certain box. No matter what.

iAdramelk:
There is two types of images used in the blog right now: Header images, avatars and images inserted using markdown should work as described already. But images inserted by using html tags by hand and skipped by `gatsby` right now. I will try to make them work too.

frontexpert:
Hi @shcheklein.
I didn't understand. Do you want to reduce the size of the image in the 50% before of process of gatsby-image that it? can you explain more to me?

fabiosantoscode:
@frontexpert I think Ivan is talking about high pixel density screens like retina, phone and 4k screens.

This is often achieved using a `<picture>` element or `srcset` image attribute with multiple sources, which can help choose the right source image for the right screen, depending on pixel density and display size.

The gatsby-remark-images plugin seems to do this out of the box. https://www.gatsbyjs.org/packages/gatsby-remark-images

The question is whether our source images are large enough.

shcheklein:
@fabiosantoscode yes, and the question is "if they are not large enough do not allow Gatsby to stretch them to the full width".

fabiosantoscode:
The set of plugins we have seem to be giving high-density images correctly to high-density screens, through a `<picture>` element. I'll create an image that's not up to par and check what happens.

fabiosantoscode:
@shcheklein I've resized an image way down to 264px and observed that the image doesn't get stretched all the way across, and instead sits at 264 *css* pixels wide. This is not retina pixels, so it's not sharp enough. However non-dense images would become tiny if shrunk to the 3x pixel density, so I don't think we should go down that road, and agree with this default behaviour.

To fix potentially small images right at the source, we can add a new check on precommit which ensures that images which have been added are at least 1400 wide and stops the commit otherwise. What do you think?

shcheklein:
@fabiosantoscode not sure I like it, I would force a smaller size (2x, 132 in your case) - just to make sure that it's always the same and it's not confusing to editors. Also, it might be an easy way out of #69 - just make original small enough.

fabiosantoscode:
@shcheklein the existing solution already yields images with always the same proportion. They're just blurrier on higher density screens :smile: 

I've done a ton of docs, issues and source code reading by now.

[...]

PR #92, opened 2020-02-02
Plugin/external resize image plugin

frontexpert:
This is solution for the issue #70

Screenshot at Feb 02 20-33-29

@shcheklein 
Please check my result and let me know your feedback.
Thanks.


shcheklein:
@frontexpert what does `title` do? is it compatible with already existing subtitle we have for images?

frontexpert:
> @frontexpert what does `title` do? is it compatible with already existing subtitle we have for images?

Yes, it is compatible, this plugin verified if have this `=200 ou =200x300 ` regex in the subtitle. If not had this make nothing and if have it gets the valor and overwrite with and or height of the image

frontexpert:
@shcheklein If you still have any problems, please let me know. Thanks.

plugins/external-rezise-image-plugin/index.js
```diff
@@ -0,0 +1,86 @@
...
+          .properties
+          .style
+          .replace(/(max-width: [0-9]{0,5}px)/g, `max-width: ${Number(width) > Number(maxWidth) ? maxWidth : width}px`);
+
+        if (!!height) {
+
+          backgroundImage.properties.style = backgroundImage
+            .properties
+            .style
+            .replace(/(padding-bottom: [0-9]{0,3}.[0-9]{0,100}%;)/g, `padding-bottom: ${height}px;`)
```
fabiosantoscode: I'm concerned to what happens to the image if the screen is resized below its `maxWidth` or `width`. The padding bottom hack is something I wouldn't use with anything but a percentage expressing the aspect ratio of the image.

plugins/external-rezise-image-plugin/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+      if (!!resize) {
+        const [width, height] = resize[0].replace(/^(.)/g, '').split('x');
+
+        image.properties.title = title
+
+        wrapperImage.properties.style = wrapperImage
+          .properties
+          .style
+          .replace(/(max-width: [0-9]{0,5}px)/g, `max-width: ${Number(width) > Number(maxWidth) ? maxWidth : width}px`);
```
fabiosantoscode: Shouldn't only this change be enough to limit the image size? I've made a change of the max-width of this element in a blog.dvc.org article and it seemed to work fine, aspect ratio and all.

fabiosantoscode:
I'm a bit concerned that this changes the existing padding-bottom hack of each image.

Also, I don't believe we need to allow the writer to set the image height. Is there a use case where we want to set the height, given the padding-bottom hack already sets the image height for us based on its real aspect ratio?

shcheklein:
> If the image subtitle is not necessary, we can write size only.

there is an already existing mechanism to put subtitles, let's not create one more and simplify code

>  > this make nothing 

could you please share the HTML is produces vs the one with size specified. `gatsby-image` regular plugin tries to create different image sizes for different screen sizes. And would like to understand how this hard coded height and width values behave on different screens, within the Gatsby pipeline, etc.

where does Gatsby takes size of images to prepare for, for example? will it understand these width and height you added?

plugins/external-rezise-image-plugin/index.js
frontexpert: This hack really exists only here to allow the writing option to set a height to the image.

plugins/external-rezise-image-plugin/index.js
frontexpert: Yes is enough

frontexpert:
> > If the image subtitle is not necessary, we can write size only.
> 
> there is an already existing mechanism to put subtitles, let's not create one more and simplify code
> 
> >  > this make nothing
> 
> could you please share the HTML is produces vs the one with size specified. `gatsby-image` regular plugin tries to create different image sizes for different screen sizes. And would like to understand how this hard coded height and width values behave on different screens, within the Gatsby pipeline, etc.
> 
> where does Gatsby takes size of images to prepare for, for example? will it understand these width and height you added?

This plugin does not change anything about the image process, it takes the image already processed by gatsby-remark-images and makes the display settings in the css.

frontexpert:
>Also, I don't believe we need to allow the writer to set the image height. Is there a use case where we want to set the height, given the padding-bottom hack already sets the image height for us based on its real aspect ratio?

I didn't think about the use cases where we want to set the height. issue #70 required to be able to put image-size(width & height) for the markdown images. I think this is not a problem, because we can omit the image height to keep the original image ratio now.

frontexpert:
Group 1

frontexpert:
@fabiosantoscode @shcheklein I removed the option of writer set height. it is more simple code now

`=WIDTH`
 
```markdown
![](/image.png)
![](/image.png "=500")
![](/image.png "=600 Some Title")
```

plugins/external-rezise-image-plugin/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "external-rezise-image-plugin",
+  "version": "1.0.0",
+  "main": "index.js",
+  "author": "Franco",
+  "license": "Apache-2.0"
+}
```
shcheklein: please add a new line

frontexpert: Ok!

fabiosantoscode: I suggest we add a .editorconfig file. A lot of editors read it natively, and it can help configure basic things like tabs/spaces, tab width and encoding, as well as whether to terminate files with a newline. I have a simple one for JavaScript projects lying around and it is consistent with what we use in this repository. Should I send a PR?

fabiosantoscode:
I'm wondering if we want to add % support here in future iterations. Or a few presets. Like `=half` to make it have a width of 50% on desktop and 100% on mobile. But I guess that's the same as simply `=350`.

gatsby-config.js
```diff
@@ -60,6 +60,7 @@ const plugins = [
             withWebp: true
           }
         },
+        'external-rezise-image-plugin',
```
fabiosantoscode: Typo: `resize`

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,63 @@
...
+        .title
+        .trim()
+        .replace(/=\d{2,4}/gi, ' ')
+        .trim();
+
+      const resize = image
+        .properties
+        .title
+        .trim()
+        .match(/=\d{2,4}/gi);
```
shcheklein: we don't need `g` here, same everywhere below

shcheklein: and above

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,63 @@
...
+    const wrapperImageList = selectAll('.gatsby-resp-image-wrapper', hast);
+
+    wrapperImageList.forEach(wrapperImage => {
+      const image = select('.gatsby-resp-image-image', wrapperImage)
+
+      const title = image
+        .properties
+        .title
+        .trim()
+        .replace(/=\d{2,4}/gi, ' ')
```
shcheklein: we don't need `i`

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,63 @@
...
+        .match(/=\d{2,4}/gi);
+
+      const maxWidth = wrapperImage
+        .properties
+        .style
+        .match(/max-width: (\d{0,5})px/gi)[0]
+        .replace(/\D/gi, ' ')
+        .trim();
+
+      if (!!resize) {
```
shcheklein: it's a bad practice to have `!!`

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,63 @@
...
+      const resize = image
+        .properties
+        .title
+        .trim()
+        .match(/=\d{2,4}/gi);
+
+      const maxWidth = wrapperImage
+        .properties
+        .style
+        .match(/max-width: (\d{0,5})px/gi)[0]
```
shcheklein: you have `i` here and not below

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,63 @@
...
+        .replace(/\D/gi, ' ')
+        .trim();
+
+      if (!!resize) {
+        const width = resize[0].replace(/^(.)/g, '');
```
shcheklein: don't need `( )` most likely, don't need `g`

shcheklein: don't need even replace for this? is there a simple option to replace first symbol? btw, are we replacing `=`? let's be explicit

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+        image.properties.title = title ? title : image.properties.alt
+
+        wrapperImage.properties.style = wrapperImage
+          .properties
+          .style
+          .replace(/(max-width: \d{0,5}px)/g, `max-width: ${Number(width) > Number(maxWidth) ? maxWidth : width}px`);
```
shcheklein: you want `min(maxWidth, width)`, right?

can it split max-width and the number into multiple lines in some cases?

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,63 @@
...
+    .use(stringify)
+    .stringify(htmlAst);
+}
+
+module.exports = ({ markdownAST }) => {
+
+  visit(markdownAST, 'html', node => {
+    const hast = convertHtmlToHast(node.value);
+
+    const wrapperImageList = selectAll('.gatsby-resp-image-wrapper', hast);
```
shcheklein: are these constants defined somewhere we can export them from? not a big deal, but might make this more stable

plugins/external-resize-image-plugin/index.js
frontexpert: Here the ` g ` is because the writer can write "=200 some title " or "some title =200" 
The is really unnecessary here

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,63 @@
+const visit = require('unist-util-visit');
+const unified = require('unified');
+const parse = require('rehype-parse');
+const stringify = require('rehype-stringify');
+const { selectAll, select } = require('hast-util-select');
+
+/** HAST - Hypertext Abstract Syntax Tree */
+function convertHtmlToHast(htmlString) {
```
shcheklein: since we use it across two plugins now - let's move them to some `plugins/utils.js`

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,63 @@
...
+        .properties
+        .style
+        .match(/max-width: (\d{0,5})px/gi)[0]
+        .replace(/\D/gi, ' ')
+        .trim();
+
+      if (!!resize) {
+        const width = resize[0].replace(/^(.)/g, '');
+
+        image.properties.title = title ? title : image.properties.alt
```
shcheklein: where does `alt` come from? I suggest we don't add additional semantics to this plugin.

plugins/external-resize-image-plugin/index.js
shcheklein: no, I think it will match it in either case w/o `g` and `i`

plugins/external-resize-image-plugin/index.js
frontexpert: OK

plugins/external-resize-image-plugin/index.js
frontexpert: By default the gatsby-remark-image already adds the alt and title in the image tag, here I am just normalizing because when we extract the resize the title ends up being null.

shcheklein: what is the `alt`  value by default? what is the `alt` value if title is provided?

frontexpert: image
When not informing title or alt the gatsby-remark-image adds by default the name of the file.
And when we set '=200' and later remove it the title ends up being null

plugins/external-resize-image-plugin/index.js
frontexpert: without ` g `
image
with ` g `
image

https://regex101.com/

plugins/external-resize-image-plugin/index.js
frontexpert: This return an array with match

```javascript
const resize = ["=200"]
```

```javascript
const resize = image
  .properties
  .title
  .trim()
  .match(/=\d{2,4}/g);
```

plugins/external-resize-image-plugin/index.js
frontexpert: Do you think is really necessary, because the const here is very specific for this plugin

plugins/external-resize-image-plugin/index.js
frontexpert: I think is better now!

fabiosantoscode:
Looks nice and simple @frontexpert :+1:

plugins/external-resize-image-plugin/index.js
frontexpert:
```javascript
 .style
 .replace(
     /(max-width: \d{0,5}px)/g,
     `max-width: ${Math.min(Number(width), Number(maxWidth))}px`
  );
```

plugins/external-resize-image-plugin/index.js
shcheklein: I mean is it defined in some gatsby global file we can import it from? so that we don't hard code the const we don't have control on (kinda depend on Gatsby internals)

plugins/external-resize-image-plugin/index.js
shcheklein: so, you are changing the logic here besides resize right? in case resize is specified (and only resize) we gonna set title to alt that would never happened unless title is specified. It changes the logic for how title is generated and is not related to image resizing. Am I missing something?

plugins/utils/convertHast.js
```diff
@@ -0,0 +1,21 @@
+const unified = require('unified');
```
shcheklein: please, update the second plugin ... that's the whole purpose of deduplicating this

plugins/external-resize-image-plugin/index.js
shcheklein: 👍 my bad, got it wrong ... please, review if we need it everywhere else, and `i`

plugins/external-resize-image-plugin/index.js
shcheklein: I understand, by why do we need ( )? what does this line do? can we avoid using regexp?

frontexpert: Yes we can!

plugins/external-resize-image-plugin/index.js
frontexpert: That's right

shcheklein: Let's avoid introducing side effects that are not related to this issue, please

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+    const wrapperImageList = selectAll('.gatsby-resp-image-wrapper', hast);
+
+    wrapperImageList.forEach(wrapperImage => {
+      const image = select('.gatsby-resp-image-image', wrapperImage)
+
+      const title = image
+        .properties
+        .title
+        .trim()
```
shcheklein: we probably don't need this trim

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+      const maxWidth = wrapperImage
+        .properties
+        .style
+        .match(/max-width: (\d{0,5})px/g)[0]
+        .replace(/\D/g, ' ')
+        .trim();
+
+      if (resize !== null) {
+        const width = resize[0].replace(/^./, '');
```
shcheklein: again, why is it `.`? should `replace('=', '')` work?

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+    const hast = convertHtmlToHast(node.value);
+
+    const wrapperImageList = selectAll('.gatsby-resp-image-wrapper', hast);
+
+    wrapperImageList.forEach(wrapperImage => {
+      const image = select('.gatsby-resp-image-image', wrapperImage)
+
+      const title = image
+        .properties
+        .title
```
shcheklein: what does it return if title is not specified at all?

plugins/external-resize-image-plugin/index.js
shcheklein: ok, I see. You are enforcing default behavior. 👍

plugins/external-resize-image-plugin/index.js
shcheklein: image name I suppose?

frontexpert: the same value of alt

plugins/external-resize-image-plugin/index.js
shcheklein: btw, just curios - why max-width is not part of those snippets ^^?

shcheklein: asking because as far as I understood the logic we expect it to be present, right?

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+        .match(/=\d{2,4}/g);
+
+      const maxWidth = wrapperImage
+        .properties
+        .style
+        .match(/max-width: (\d{0,5})px/g)[0]
+        .replace(/\D/g, ' ')
+        .trim();
+
+      if (resize !== null) {
```
shcheklein: it should be the same as just `if (resize)`, no?

plugins/external-resize-image-plugin/index.js
frontexpert: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-images/src/index.js#L179

this is the ref the gatsby-remark-image

plugins/external-resize-image-plugin/index.js
frontexpert: Yes,

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+        const width = resize[0].replace(/^./, '');
+
+        image.properties.title = title ? title : image.properties.alt
+
+        wrapperImage.properties.style = wrapperImage
+          .properties
+          .style
+          .replace(
+            /(max-width: \d{0,5}px)/g,
```
shcheklein: you don't need `( )`? also, does it always come in the style block like this - with one space?

should it be also `\d{1,5}`?

plugins/external-resize-image-plugin/index.js
shcheklein: 👍

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+            `max-width: ${Math.min(Number(width), Number(maxWidth))}px`
```
shcheklein: probably you don't need to apply Number here, btw ... most likely it can convert it itself

plugins/external-resize-image-plugin/index.js
frontexpert: That's right

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+      const resize = image
+        .properties
+        .title
+        .trim()
+        .match(/=\d{2,4}/g);
+
+      const maxWidth = wrapperImage
+        .properties
+        .style
+        .match(/max-width: (\d{0,5})px/g)[0]
```
shcheklein: same here ... since it's the same (almost) the same regexp as below - move into const to avoid duplication

same question - does it always come with `max-width`?

shcheklein: should it be `[1]`, btw?

frontexpert: Here I get the value of max-width of wapper of the image that is tag span
image

shcheklein: I see, thanks. Other comments still apply probably. Would great to put some comment before select all that describes this span and img relationship.

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+        .match(regexResize);
+
+      const maxWidth = wrapperImage
+        .properties
+        .style
+        .match(regexMaxWidth)[0]
+        .replace(/\D/g, '')
+
+      if (resize) {
+        const width = resize[0].replace(/=/, '');
```
shcheklein: you don't need regexp ... just `replace('=', '')` should do the trick .. i think

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+        image.properties.title = title ? title : image.properties.alt
```
shcheklein: let put a comment here:

`# by default Gatsby populates title value with alt, restoring it here if needed`

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+const regexMaxWidth = /max-width: \d{1,5}px/g
+const regexResize = /=\d{2,4}/g
+
+module.exports = ({ markdownAST }) => {
+
+  visit(markdownAST, 'html', node => {
+    const hast = convertHtmlToHast(node.value);
+
+    const wrapperImageList = selectAll('.gatsby-resp-image-wrapper', hast);
```
shcheklein: let's add a comment here:

```
# image related HTML produced by Gatsby looks like: 
# <span .gatsby-resp-image-wrapper max-width: 100px>
#     <img .gatsby-resp-image-image title='..' alt='...' >
#...
```

make it right ^^ it's just a draft so you get the idea

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,52 @@
+const visit = require('unist-util-visit');
```
shcheklein: put a comment here that describes the markdown syntax this plugin supports

plugins/external-resize-image-plugin/index.js
```diff
@@ -34,7 +43,16 @@ module.exports = ({ markdownAST }) => {
         .replace(/\D/g, '')
 
       if (resize) {
-        const width = resize[0].replace(/=/, '');
+
+        const width = resize[0].replace('=', '');
+
+        //# by default Gatsby populates title value with alt,
```
shcheklein: sorry, I didn't mean to put `#` before each string :) it was a designation for a single-line comment  - `//` is def enough

plugins/external-resize-image-plugin/index.js
```diff
@@ -34,7 +43,16 @@ module.exports = ({ markdownAST }) => {
...
+
+        const width = resize[0].replace('=', '');
+
+        //# by default Gatsby populates title value with alt, 
+        //# restoring it here if needed
+
+        //# image related HTML produced by Gatsby looks like: 
+        //# <span .gatsby-resp-image-wrapper max-width: 100px>
+        //#   <img .gatsby-resp-image-image title='..' alt='...'>
+        //# ...
```
shcheklein: this comment should go before `selectAll`

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,75 @@
...
+//# ![](/relative-path-image "=500 Some Title")
+//# ![](/relative-path-image "Some Title =500")
+
+const visit = require('unist-util-visit');
+const { selectAll, select } = require('hast-util-select');
+
+const { convertHastToHtml, convertHtmlToHast } = require('../utils/convertHast');
+
+const regexMaxWidth = /max-width: \d{1,5}px/g
+const regexResize = /=\d{2,4}/g
```
shcheklein: move these consts inside appropriate methods. No need to keep them global. It's easier to read when they are close.

plugins/external-resize-image-plugin/index.js
```diff
@@ -0,0 +1,75 @@
...
+        //# restoring it here if needed
+
+        //  image related HTML produced by Gatsby looks like: 
+        //  <span .gatsby-resp-image-wrapper max-width: 100px>
+        //    <a .gatsby-resp-image-link href='/static/...'>
+        //      <span .gatsby-resp-image-background-image background-Image>
+        //      <picture>
+        //        <source srcset="/static/..">
+        //        <source srcset="/static/..">
+        //        <img .gatsby-resp-image-image title='..' alt='...' max-width: 100%>
```
shcheklein: move this comment before selectAll or close to that one. this place is not relevant for it

plugins/external-resize-image-plugin/index.js
```diff
@@ -1,20 +1,33 @@
...
-//# ![](/relative-path-image "Some Title =500")
+// Support for resize image inline on markdown
+// Syntax "=WIDTH" 
+//
+// Examples
+// ![](/relative-path-image "=500")
+// ![](/relative-path-image "=500 Some Title")
+// ![](/relative-path-image "Some Title =500")
+
+//  by default Gatsby populates title value with alt,
```
shcheklein: @frontexpert why do you put all comments in one place? each of those is relevant to a specific scope. First one is global, second one is to the place where you update title, last one close to selectAll and everything after that

plugins/external-resize-image-plugin/index.js
shcheklein: ok, looks like the last one ☝️ any ideas, does Gatsby export these names by any chance?

plugins/external-resize-image-plugin/index.js
frontexpert: I fixed it

shcheklein: thanks!

plugins/external-resize-image-plugin/index.js
frontexpert: In the gatsby-remark-images have a file that exports these class CSS
image

shcheklein: so, let's import them, instead of hardcoding!

frontexpert: OK!

frontexpert:
@shcheklein 
which issues/tickets can I work on next? I want you to assign me new issues/tickets.

shcheklein:
@frontexpert now that it's merged, let's use it to solve #40 ? we need to go to blog.dataversioncontrol.com and get as higher res as possible for all images (mostly older blog posts). See how medium renders and them and make smaller those that look too blurry.

PR #93, opened 2020-02-03
images: store images in DVC

fabiosantoscode:
After many iterations, and many things cleared up by @shcheklein, I'm pretty confident this makes sense.

The hardest part was preventing an infinitely updating process, in which a pull caused a push through file change events, which then caused a pull and so on. I tried many ways of avoiding this, some of which had semi-decent results, but none were as readable as the simple loop that I've ended up with.

Right now, you can add, remove and change images, and this causes a `dvc commit -f` followed by a `dvc push`. If you change the DVC file (or switch branches) this causes a `dvc pull`. If any of these commands fails (because of changes you might not want to overwrite), the plugin crashes gatsby (through a `reporter.panic` call) and shows you the command that failed, the stderr output and exit code.

NOTE: This PR comes with a remote only I have access to, and we need to figure out where to place the images.


fabiosantoscode:
Got netlify to install DVC (build as a whole is currently failing because it lacks credentials, but it gets to `gatsby buid`)

fabiosantoscode:
Also, I found out about this official gatsby plugin today: https://www.gatsbyjs.org/packages/gatsby-plugin-netlify-cache/. It caches gatsby's stuff in /opt/build/cache, netlify's cache directory. It also accepts extra cache directories so we can cache `.dvc/cache` there as well. Less S3 traffic to pay for!

Want me to add this as part of this PR? I didn't because I feel it's complex enough.

.dvc/config
```diff
@@ -0,0 +1,4 @@
+['remote "s3"']
+url = s3://fabio.images.dvc.org
```
shcheklein: the way you usually do it is:

`dvc remote add -d storage https://.......` (https read-only access to S3 - for services like Netlify to have access)
`dvc remote add -d --local storage s3://fabio.images.dvc.org` - overrides locally your settings to being able to publish changes

fabiosantoscode: Sounds good!

fabiosantoscode:
@shcheklein if we can tell which files have changed (either by using the DVC API directly or introducing detailed JSON output to `dvc status`), we can ask the user which version they want to keep, for every file which they changed without syncing to DVC or running gatsby develop.

You expressed concern about an error occurring between `dvc commit -f` and `dvc push`. I've given it some thought and to me it seems that it's fine as long as the user gets an error message. If `dvc commit` errors, `gatsby develop` will crash with the error message and the user will be given a chance to try again (by restarting gatsby develop).

Probably a more gracious way of telling the user would be good, like only crashing if it's a `gatsby build` but in `gatsby develop` we hot-reload an error overlay with a retry button.

shcheklein:
let's rebase? 

> if we can tell which files have changed (either by using the DVC API directly or introducing detailed JSON output to dvc status), we can ask the user which version they want to keep, for every file which they changed without syncing to DVC or running gatsby develop.

the lates version of `dvc diff` does this to some extent and has json output - could you try it and give me some feedback?

> gatsby develop will crash with the error message and the user will be given a chance

kk

requirements.txt
```diff
@@ -0,0 +1,51 @@
+appdirs==1.4.3
```
shcheklein: any luck with a proper packaging?

plugins/dvc-image-plugin/index.js
```diff
@@ -0,0 +1 @@
+//no-op
```
shcheklein: why don't we use this file (vs gatsby-node.js)? is it a convention of some sort?

plugins/dvc-image-plugin/gatsby-node.js
```diff
@@ -0,0 +1,76 @@
+const { spawn } = require('child_process');
+const chokidar = require('chokidar');
+
+const imagesDvcPath = 'uploads.dvc';
```
shcheklein: so, what about configuring this via some config in gatsby?

fabiosantoscode: I was leaving that for a future iteration. If you're feeling good about where this is going, I'll get right on it.

plugins/dvc-image-plugin/index.js
fabiosantoscode: The docs recommend that an `index.js` file is created when there is no `main`. They don't mention why, but one advantage of this is that the module can be discovered by `require.resolve`, so that might be it.

`gatsby-node.js` has a special meaning for gatsby, so if we placed its code in this file, gatsby wouldn't call the hooks we export.

Reference: https://www.gatsbyjs.org/docs/files-gatsby-looks-for-in-a-plugin/

requirements.txt
fabiosantoscode: Not in netlify. There's no sudo, therefore no apt.

Unless we distribute an AppImage (doesn't need sudo to install, it's a standard executable file), a distro-specific executable, or an npm-wrapped executable, we can't really install without sudo.

One thing we can do is add a single line to the requirements saying we want `dvc==0.82.*`, that should force `pip` to fetch the rest of the dependencies without us specifying or versioning them.

fabiosantoscode:
I had an idea to make life easier for contributors who don't have S3 access, and also for contributors with S3 access who don't want to add S3 credentials to their environment when pushing (I need to prefix my AWS-enabled commands with `aws-personal.sh ...`, for example. Others might `source ~/path/to/creds.sh`).

We can tell them to place their created images in `static/uploads` and track them with git using `git add -f static/uploads/2020-01-01-foo.png`.

Using a PR github action, we can look for these non-DVC tracked files using `git ls-files static/uploads/`. Then proceed to remove them from the repository, track them in git, commit the result to master and close the PR. This is a way to enable non-contributors to push new data to DVC in a secure way, not only on this project but on others as well.

Alternatively, we can use a CircleCI step or github action to take care of the commit right when it lands on master. We can force-push to avoid the images from making our `.git` folder larger over time.

requirements.txt
shcheklein: moving forward we need to package this as an npm package, so it not worth effort iterating on pip, apt and other stuff ... we can start doing research in the right direction

fabiosantoscode:
@shcheklein I know I wrote a wall of text (again), but please give some thought to my comment above. We can even get rid of this PR entirely and make this a pre-build step in a simple node or bash script.

shcheklein:
@fabiosantoscode nothing wrong with your comment :) I just don't have enough time to review and give it a thought. I'll get to it.

PR #95, opened 2020-02-05
images: wrap text around images

### LINKED ISSUE #71: option for text wrapping
It would be great to have an option to wrap text, as is that doesn't seem to be doable by manipulating the html  parameters. 

image

fabiosantoscode:
Also added some tests to make sure both features don't collide. More tests can be added here by doing a simple function extraction on the plugin.

Fixes #71


shcheklein:
@fabiosantoscode please rebase, I've pushed minor style changes (applied prettier and lint)

plugins/external-resize-image-plugin/index.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+describe('extractInstructions', () => {
+  it('extracts the title if no instructions are found', () => {
+    expect(extractInstructions('I am a title')).toEqual({
+      title: 'I am a title',
+      resize: null,
+      wrap: null
+    });
+  });
+  it('extracts a resize instruction when it finds =NNN', () => {
```
shcheklein: again, please let's add a new line in between - it's easier to read and it is the way it's done everywhere else

plugins/external-resize-image-plugin/index.js
```diff
@@ -1,5 +1,5 @@
 // Support for resize image inline on markdown
-// Syntax "=WIDTH"
```
shcheklein: how does syntax for wrap look like?

src/components/post/styles.module.css
```diff
@@ -211,3 +211,28 @@
     opacity: 0.7;
   }
 }
+
+:global {
+  .gatsby-resp-image-wrapper {
```
shcheklein: is there a way to import it? I mean the const name

plugins/external-resize-image-plugin/index.js
```diff
@@ -44,33 +59,53 @@ module.exports = ({ markdownAST }) => {
...
+      const maxWidth = wrapperImage.properties.style
         .match(regexMaxWidth)[0]
-        .replace(/\D/g, '')
+        .replace(/\D/g, '');
 
-      if (resize) {
+      if (resize || wrap) {
+        //  by default Gatsby populates title value with alt,
+        //  restoring it here if needed
+        image.properties.title = image.properties.alt = title;
```
shcheklein: this does not look right to me ... look like we are overriding alt with title, even if it was specified

plugins/external-resize-image-plugin/index.js
```diff
@@ -8,32 +8,47 @@
...
-  const regexResize = /=\d{2,4}/g
-
-  const title = string.replace(regexResize, '').trim()
+const {
+  imageClass,
+  imageWrapperClass
+} = require('gatsby-remark-images/constants');
+const {
+  siteMaxWidth,
+  imageWrapClassPrefix,
```
shcheklein: don't see any reason to introduce global class consts in this case, tbh

and we should have site width already defined somehwere ...

plugins/external-resize-image-plugin/index.js
```diff
@@ -44,33 +59,53 @@ module.exports = ({ markdownAST }) => {
...
+          regexMaxWidth,
+          `max-width: ${Math.min(resize, maxWidth)}px`
+        );
       }
-    })
-    node.value = convertHastToHtml(hast)
-  })
+    });
+
+    const stopWrapTagList = selectAll('stop-wrap', hast);
```
shcheklein: why do we need this?

src/components/post/styles.module.css
fabiosantoscode: Yes, I'll try this: https://github.com/css-modules/icss#import I have to create a CSS file with it though, can't be javascript :(

plugins/external-resize-image-plugin/index.js
fabiosantoscode: They are both the same in an image tag:

`altAndTitle`

Currently the alt attributes of an image with a specific width have `=200` in it.

Maybe I could strip the alt and the title separately but they seem to be exactly the same. Alt for blind people, and title for people hovering with the mouse.

plugins/external-resize-image-plugin/index.js
fabiosantoscode: A quick `git grep 700` revealed that there was no constant.

Do you suggest that the imageWrapClassPrefix const be placed in the plugin folder?

plugins/external-resize-image-plugin/index.js
fabiosantoscode: To provide a clearfix when you don't want to wrap anymore. For example, you might want only 2 paragraphs wrapping around the image. After those, you place a `<stop-wrap />` and the paragraphs after it will follow below the image. It's just a div with `clear:both`

plugins/external-resize-image-plugin/index.js
fabiosantoscode: `:wrap-left` and `:wrap-right`. I'll add to the comment.

plugins/external-resize-image-plugin/index.test.js
fabiosantoscode: Sorry!

src/components/post/styles.module.css
fabiosantoscode: Looks like importing classes with this is not possible. I looked at postcss modules which might do this but they're all unmaintained at this point :/

plugins/external-resize-image-plugin/index.js
fabiosantoscode: This screenshot illustrates what the clearfix is for: the left-wrapped image only has a paragraph to its left. The paragraph after it is below the image due to such a clearfix. The right-wrapped image doesn't have a clearfix, so even the title wraps around its left.

Screenshot from 2020-02-05 15-09-34

plugins/external-resize-image-plugin/index.js
fabiosantoscode: I figured you did want me to move the image-specific constants to the plugin so I did that. Can always revert.

plugins/external-resize-image-plugin/index.js
shcheklein: so, what happens if you do:

`'some alt here'`?

plugins/external-resize-image-plugin/index.test.js
shcheklein: no worries :) I was using "again" in a sense of "same here"

src/components/post/styles.module.css
shcheklein: @iAdramelk can postcss do this? import class names to JS and CSS so that you have them defined in one place?

shcheklein:
@fabiosantoscode just a few minor questions to clarify, otherwise looks great!

plugins/external-resize-image-plugin/index.js
fabiosantoscode: It would have no effect. Just like before, you would have to add the instructions to the title.

shcheklein: @fabiosantoscode I didn't get your answer, sorry :( so, what would it generate in this case? ('=200' - is the title).

fabiosantoscode: I'd never seen that syntax before :) I'll change this right away!

fabiosantoscode: The reason this change is here is because I was using the alt specifier instead of the title specifier.

gatsby-config.js
```diff
@@ -56,7 +57,7 @@ const plugins = [
         {
           resolve: 'gatsby-remark-images',
           options: {
-            maxWidth: 700,
+            maxWidth: siteMaxWidth,
```
shcheklein: probably better name imageMaxWidth or siteImageMaxWidth ... current name is not exactly correct

shcheklein:
@fabiosantoscode check also, there are some deployment problems it looks like

gatsby-config.js
fabiosantoscode: Maybe `contentMaxWidth` since this is the maximum width of a paragraph of content too, not just an image.

fabiosantoscode:
Renamed the constant @shcheklein, and the build seems fine rn.

shcheklein:
@fabiosantoscode it's not content though, it looks to me like it affects only images (it's an option passed to gatsby-remark-plugin). Am I missing something? 🤔

fabiosantoscode:
@shcheklein If you measure the width of a paragraph or title you'll see they also turn out to be 700 wide on desktop displays. It's the width of the body of the post, be it text or images.

shcheklein:
@fabiosantoscode makes sense, but then you should start using it everywhere explicitly instead of hardcoded 700px. We can name it as it was before in this case. The way it is used now across the code and its name is misleading.

fabiosantoscode:
My intention was to start using it everywhere indeed. But I'll revert the name.

shcheklein:
@fabiosantoscode and let's start using it, or otherwise let's now introduce it all. It's just makes it super confusing - when you see a generic name "contentWidth" but it affects only images.

fabiosantoscode:
I see what you mean. It just so happens that the rest of the usages of `700` are all in CSS, and if we want a CSS constant we need to declare it in CSS.

src/constants.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
shcheklein: I wonder if it should be part of the gatsby config? Is there a convention like this to store that kind of custom configs?

plugins/external-resize-image-plugin/index.js
```diff
@@ -15,28 +17,41 @@ const {
...
+
 const {
   convertHtmlToHast,
   convertHastToHtml
 } = require('../utils/convertHast');
 
-const extractResize = str => {
+const extractInstructions = titleString => {
   const regexResize = /=\d{2,4}/g;
+  const regexWrap = /:wrap-(left|right)/;
```
shcheklein: do we need a `g` here?

shcheklein:
> I see what you mean. It just so happens that the rest of the usages of 700 are all in CSS, and if we want a CSS constant we need to declare it in CSS.

yep, thanks

jest.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  testPathIgnorePatterns: ['.cache']
```
shcheklein: do we need both - this and jestignore?

plugins/external-resize-image-plugin/index.js
```diff
@@ -1,11 +1,13 @@
 /*
```
shcheklein: while you are on this - could we rename the plugin ... `external` is not needed actually

fabiosantoscode: Sure!

jest.config.js
fabiosantoscode: Great point. I think we don't. I'm deleting .jestignore.

plugins/external-resize-image-plugin/index.js
fabiosantoscode: We could add it, just so people can type `:wrap-left :wrap-left` by mistake and one of them doesn't end up in the title.

shcheklein: I don't remember exact semantics of `g` in JS. I just see that we use it a line above. I wonder if it's about _all_ or about _match globally_ (we must have `g` in this case then).

Either way it's probably either bug here or a line above.

fabiosantoscode: Actually, I can't because it changes the behaviour of the capture group `(left|right)`. It would start returning `:wrap-left` and `:wrap-right` as a result and I'd have to coax the `left` or `right` part out of each. Pretty convoluted for a simple match.

src/constants.js
fabiosantoscode: I had a look at the docs and some of the sites on the gatsby github account, and didn't find any site-wide constants. The docs do say that a gatsby project is a react project and inherits a lot of react conventions. However, I've never seen anyone agree on where to store site-wide constants. `src/config` and `src/constants` are some filenames I've seen.

plugins/external-resize-image-plugin/index.js
shcheklein: k, it's fine. np. I would just probably reconsider this behavior for `=xxx` expressions as well then. Just capture the first one. To make it consistent. I think it does not make sense to have multiple of those.

src/constants.js
shcheklein: 👍

PR #98, opened 2020-02-06
add hero component

frontexpert:
#40 
This component doesn't stretch image when is smaller than max-width

Side left with component and side right without component

image

image

image

### DIFF
--- a/src/components/hero/index.tsx
+++ b/src/components/hero/index.tsx
@@ -1,0 +1,72 @@
+import Image from 'gatsby-image';
+import React from 'react';
+
+import { IBlogPostFrontmatter } from '../../templates/blog-post';
+
+import styles from './styles.module.css';
+
+interface IFluidObject {
+  aspectRatio: number;
+  src: string;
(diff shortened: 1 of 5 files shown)


frontexpert:
@shcheklein Please check my PR and let me know your feedback.
Thanks.

shcheklein:
@frontexpert looks cool 😎 I love it :) but it's an overkill again, sorry.

How many posts do we have where hero image does not have a proper size? 2? Let's just find or make with canva something 2000px and replace them. We won't need this sophisticated logic in the future (as you could tell by some recent posts).

What should we do instead (to repeat myself):

1. we should never allow any stretching, we should actually should even make 1/2 px (show 350px if the origin is 700px) - that would be a sign for the editor to replace/make a better image.
2. we should go through all (not only hero) images in the older blog posts and either apply the logic above ^^, or find on medium better origins or apply `=size` to render them smaller.

frontexpert:
I'm closing this PR to #99

PR #99, opened 2020-02-07
add sufficiently small original image size

### LINKED ISSUE #40: properly size post images

frontexpert:
Fixes #40 
I think it is enough now! #98  

Now I added the max-width default in a constant and in the plugin I take the width of the original image and multiply for 2 and compare with his constant if it is smaller than constant I divide for 2 and add in max-width if not I make nothing

Now if the writer adds in a post an image with 700px the Gatsby is show an image of 350px

before  /  after
image

Before, the original image has 614px and show the original size of 614px
image

After, the original image continue with 614px but now show 307px in the max-width
image


frontexpert:
@shcheklein Please check my result again, and let me know your feedback.
Thanks.

fabiosantoscode:
@shcheklein on this screenshot you can see the problem I mentioned before. The text became too small. Is it OK to have tiny text just to avoid it being blurry? It looks sharper, but decreases readability.

shcheklein:
@fabiosantoscode it's okay. These are older images, I don't worry that much about them. @frontexpert I would check though if we can get batter originals from medium (by playing with URL).

@frontexpert what about two or three older hero images? we need to fix them (it's fine to find some replacements or use canva)

@frontexpert what about hero block in general? we should be applying the same logic.

also, keep in mind @fabiosantoscode 's PR #71 - it'll be merged before and you will need to adjust your code.

frontexpert:
I used this https://deep-image.ai/ tool for enhancing images

fabiosantoscode:
That's pretty nifty!

frontexpert:
preview here https://issue-40.netlify.com/ml-best-practices-in-pytorch-dev-conf-2018

frontexpert:
@shcheklein Did you check my state?

shcheklein:
@frontexpert check this comment - https://github.com/iterative/blog/pull/99#issuecomment-583504631 . I think not all the issues/question were addressed here.

Check you PR carefully and cleanup - unused images, etc. Please make one commit and force push it to avoid saving all these images we don't need into Git history.

fabiosantoscode:
Re pushing a single commit, it's possible to set squash+merge as the default merge button action on a GitHub repository. You get to create a commit message when you merge, and it includes the PR number by default, for later auditing.

shcheklein:
@frontexpert @fabiosantoscode yep, I know about the default option to squash (it's not set yet, since I don't always like it tbh), but probably for this specific repo it makes the most sense.

frontexpert:
I am closed for PR  #103 
clean commit

PR #100, opened 2020-02-07
ci: add CircleCI integration and ci-check script.

### LINKED ISSUE #10: add CI integration and run format checker, linter, etc

fabiosantoscode:
I haven't fixed the linting or typing errors so I've commented out that part.

Also added a `yarn check-all` so we can run `prettier --check` instead of `prettier --write` to ensure all our files are merged to master properly linted.

Fixes #10

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,55 @@
+# Javascript Node CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/language-javascript/ for more details
+#
+version: 2.1
+
+defaults: &defaults
+  working_directory: ~/repo
+  docker:
+    # specify the version you desire here
+    - image: circleci/node:10.16.3
+
+    # Specify service dependencies here if necessary
+    # CircleCI maintains a library of pre-built images
+    # documented at https://circleci.com/docs/2.0/circleci-images/
+    # - image: circleci/mongo:3.4.4
+
+commands:
+  install:
+    steps:
+      - checkout
+
+      # Download cached dependencies
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "yarn.lock" }}
+            # fallback to using the latest cache if no exact match is found
+            - v1-dependencies-
+
+  cache:
+    steps:
+      # Upload dependencies cache
+      - save_cache:
+          paths:
+            - node_modules
+          key: v1-dependencies-{{ checksum "yarn.lock" }}
+jobs:
+  lint:
+    <<: *defaults
+    steps:
+      - install
+      - run: yarn typecheck
+      - run: yarn check-all
+      - run: yarn lint-ts
+      - run: yarn lint-css
+      - cache
+
+workflows:
+  version: 2
+
+  commit:
+    jobs:
+      - test
+        filters:
+          branches: { ignore: master }
--- a/package.json
+++ b/package.json
@@ -9,15 +9,16 @@
     "develop": "gatsby develop",
     "serve": "gatsby serve",
     "start": "yarn develop",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "tsc --noEmit --skipLibCheck",
     "test": "jest",
     "lint-ts": "tslint --project .",
     "lint-css": "stylelint \"src/**/*.css\"",
-    "format-all": "prettier --write './**/*.{js,jsx,md,tsx,ts,json}'"
+    "format-all": "prettier --write './**/*.{js,jsx,md,tsx,ts,json}'",
+    "check-all": "prettier --check './**/*.{js,jsx,md,tsx,ts,json}'"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged && yarn typecheck"
     }
   },
   "lint-staged": {
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+.cache/
+public/
--- a/config/postcss/media.js
+++ b/config/postcss/media.js
@@ -1,3 +1,5 @@
+/* tslint:disable object-literal-sort-keys */
+
 const screens = {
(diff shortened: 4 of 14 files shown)


fabiosantoscode:
@shcheklein I don't have permissions but we probably need to enable CircleCI for this repository.

.circleci/config.yml
```diff
@@ -0,0 +1,50 @@
...
+      - install
+      - run: yarn ci-check
+      - cache
+
+workflows:
+  version: 2
+
+  commit:
+    jobs:
+      - test
```
fabiosantoscode: This is largely a copy of what we have in the dvc.org repo, trimmed down to what we need.

.circleci/config.yml
```diff
@@ -0,0 +1,50 @@
+# Javascript Node CircleCI 2.0 configuration file
```
shcheklein: I think you need more stuff from the dvc.org's file - restoring the branch for example.

fabiosantoscode: I saw that it was a workaround for tools which compare HEAD to master. None of the tools we're using do that.

package.json
```diff
@@ -12,7 +12,9 @@
     "typecheck": "tsc --noEmit",
     "lint-ts": "tslint --project .",
     "lint-css": "stylelint \"src/**/*.css\"",
-    "format-all": "prettier --write './**/*.{js,jsx,md,tsx,ts,json}'"
+    "format-all": "prettier --write './**/*.{js,jsx,md,tsx,ts,json}'",
+    "check-all": "prettier --check './**/*.{js,jsx,md,tsx,ts,json}'",
+    "ci-check": "yarn check-all && yarn lint-css # TODO lint and type files && yarn lint-css && yarn typecheck"
```
shcheklein: please, don't include TODO - create a ticket if there is something to be fixed

shcheklein: we should not run the full check - it's slower and is not related to PR,  I think we always run only files from the PR

shcheklein: and yep, we need all other checks we have

fabiosantoscode: Got it

fabiosantoscode: The checkers we are using are pretty fast. But I'll try to check only modified files.

fabiosantoscode:
Ok, now only files which have changed between master and the PR will be linted. I didn't find any package which did this out of the box so I made a script for it. And because we're already configuring `lint-staged` I thought I might as well read its configuration directly.

The script is essentially `git diff master --name-only | xargs <linter>`, and selects which linter to use according to file globs.

The output looks like this:

```
LINT-CHANGED> yarn prettier --write scripts/lint-changed.js
yarn run v1.17.3
$ /home/fabio/devel/iterative/blog/node_modules/.bin/prettier --write scripts/lint-changed.js
scripts/lint-changed.js 59ms
Done in 0.26s.

LINT-CHANGED> yarn lint-ts scripts/lint-changed.js
yarn run v1.17.3
$ tslint --project . scripts/lint-changed.js
Done in 1.89s.
```

shcheklein:
@fabiosantoscode what are the checks on the dvc.org that required clearing the master? Trying to understand the logic behind stuff. Current solution looks too complicated - and there are a few ways to make it simpler, but first I'd like to restore the full picture of what is going on on dvc.org and why.

shcheklein:
so, looking closely into this I think we don't need this restore master procedure anymore (after @jorgeorpinel changed it from pretty-quick to the direct prettier call.

let's confirm that dvc.org actually runs all the checks (except links) for the whole project and then cleanup that master branch restoration on both projects and get back to your previous simple logic here. Except that we need to run all checks - the same way we do for dvc.org. Sorry for the confusion around checks on the dvc.org end.

fabiosantoscode:
@shcheklein ok, I'll go back to that.

No worries, it should be really fast - many times faster than the netlify deploy - even if we check all files.

I'll revert the changes.

Do you think you need to enable circleCI in this repository? I'd like to see if this actually runs there.

fabiosantoscode:
What's happening on dvc.org is a full repo check every time.

fabiosantoscode:
@shcheklein I suggest we do the full-project linting as part of https://github.com/iterative/blog/issues/101. It also includes adding the missing checks where I had that TODO comment before. But I can do it in this PR if you prefer.

package.json
```diff
@@ -12,7 +12,9 @@
     "typecheck": "tsc --noEmit",
     "lint-ts": "tslint --project .",
     "lint-css": "stylelint \"src/**/*.css\"",
-    "format-all": "prettier --write './**/*.{js,jsx,md,tsx,ts,json}'"
+    "format-all": "prettier --write './**/*.{js,jsx,md,tsx,ts,json}'",
+    "check-all": "prettier --check './**/*.{js,jsx,md,tsx,ts,json}'",
+    "ci-check": "yarn check-all && yarn lint-css"
```
shcheklein: it does not look like _all_?  we have lint-ts ... (and yes, let's fix the complaints)

also, it's better to keep steps in the CircleCi separate - easier to navigate and see what specific check fails and you can avoid introducing this "ci-check" here - it's not needed

shcheklein:
please, rebase - there is a conflict

shcheklein:
I've enabled CircleCI, but I'm not sure I can it for the fork

shcheklein:
I suppose we'll have to merge it and see how it goes

fabiosantoscode:
@shcheklein squashed it up and added the typecheck before each commit. Typechecking individual files didn't work but the full project check is less than 3 seconds.

shcheklein:
@fabiosantoscode what is needed for? I mean typecheck. I'm not familiar but don't we run `tsc` to compile code anyway?

fabiosantoscode:
@shcheklein the typecheck is for verifying that the types check out. For example, it checks that you're not passing numbers to a function that takes strings.

shcheklein:
@fabiosantoscode merged! check the results here - https://circleci.com/gh/iterative/blog/2

PR #102, opened 2020-02-07
tags: create tag lists to list posts of a certain tag

### LINKED ISSUE #46: make tag names clickable: show description and posts in the tag

fabiosantoscode:
I've had to use GraphQL fragments to reuse queries. I've used fragments before and they were very effective when reusing components which need data. Gatsby's integration seems top notch. My reasoning for this was that the home page and the tag page use the same Feed component, and need to provide data to it without duplicating the complex Feed query.

I also think we could use centralisation in our types. Structures like `{ edges: Array<{ node: SomeType }> }` needn't be repeated.

Let me know if you'd like pagination in the tag pages. I don't think we'll need it for a long time.

Fixes #46

### DIFF
--- a/src/components/feed/index.tsx
+++ b/src/components/feed/index.tsx
@@ -1,4 +1,8 @@
+import { graphql } from 'gatsby';
+
 import React from 'react';
+
+import cn from 'classnames';
 
 import { IFeedPostData } from '../../pages/index';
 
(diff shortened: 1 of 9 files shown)


fabiosantoscode:
Here's what the tag page looks like. It's just a `<Feed />` with some props I've added.

Screenshot from 2020-02-07 18-29-27

This is probably not a good color for clickable links, and also there's no hover or focus styling. But I don't want to go down the design road, because I'm really bad at it.

Screenshot from 2020-02-07 18-29-51

fabiosantoscode:
Seems like there's some `gatsby build` issues. I'll get to this on Monday.

fabiosantoscode:
The gatsby build issue is an import ordering/conflict thing. I'd think this didn't happen with CSS modules and I find this pretty weird.

fabiosantoscode:
There we go. I just had to ask the webpack CSS chunk plugin (which doesn't know our CSS files are CSS modules) to ignore the ordering.

fabiosantoscode:
@shcheklein I've fixed the build issues this morning, care to have a look?

shcheklein:
@fabiosantoscode I'll do this asap! Just haven't had enough time today, sorry :(

fabiosantoscode:
Take your time. I'll be in bed soon :)

src/components/feed/index.tsx
```diff
@@ -7,31 +9,45 @@ import FeedItem from './item';
...
-    {
-      node: IFeedPostData;
-    }
-  ];
+  feedPostList: Array<{
+    node: IFeedPostData;
+  }>;
+  bigFirst?: boolean;
+  header?: React.ReactNode;
+  leadParagraph?: React.ReactNode;
```
iAdramelk: Right now `header` and `leadParagraph` are marked as optional in types, but there is no check for their existence down in the templates. We should either make them required or remove blank div tags in template below if they are not set.

fabiosantoscode: I didn't touch this code, this was a move from pages/index

fabiosantoscode: But I'll make this required.

iAdramelk: Thank you!

src/utils/tag-to-slug.js
```diff
@@ -0,0 +1,5 @@
+module.exports = tag =>
+  tag
+    .toLowerCase()
```
shcheklein: probably better to run trim (or strip or how is called in JS?)

shcheklein:
can we add just a little bit of responsiveness to the buttons - slightly change shade of grey would be enough?

shcheklein:
tests fail btw, for some reason

src/utils/tag-to-slug.js
fabiosantoscode: Good call.

fabiosantoscode:
It's the types :) I'm fixing that atm.

fabiosantoscode:
I'll go get that hover/focus state on the buttons after dinner. Tests should pass now.

fabiosantoscode:
Took advantage of `--color-gray-hover`. It's a darker shade of gray.

Screenshot from 2020-02-12 19-50-18

fabiosantoscode:
@shcheklein when you said responsiveness, I hope you meant responsiveness to hover and focus :)

fabiosantoscode:
Removed a header max-width that was meant for the home page, from the tag page.

Screenshot from 2020-02-12 20-57-57

(Also why is that not an `h2`? Any specific reason?)

shcheklein:
> (Also why is that not an h2? Any specific reason?)

no, good reason. Probably should be h2.

> I hope you meant responsiveness to hover and focus :)

yes, and your solution looks great.

fabiosantoscode:
Changed to a h2 and squashed :)

shcheklein:
I wonder why is it now deploying it for you ? 🤔

shcheklein:
ok, deployment is running now

fabiosantoscode:
The circle job runs first, then deployment. Not sure why.

PR #103, opened 2020-02-10
Fix:  size post images issue 40

### LINKED ISSUE #40: properly size post images

frontexpert:
Fixes #40


gatsby-config.js
```diff
@@ -2,6 +2,7 @@ const path = require('path');
 require('./config/prismjs/dvc');
 const { imageMaxWidth } = require('./src/constants');
 
+
```
shcheklein: ?

plugins/resize-image-plugin/index.js
```diff
@@ -77,6 +78,11 @@ module.exports = ({ markdownAST }) => {
         image.properties.title = title ? title : image.properties.alt;
       }
 
+      const originalSize = source.properties.srcSet
```
shcheklein: it's unused?

src/templates/blog-post.tsx
```diff
@@ -107,8 +107,9 @@ export const pageQuery = graphql`
         }
         picture {
           childImageSharp {
-            fluid(maxWidth: 850, cropFocus: CENTER) {
```
shcheklein: ? why do we remove this?

shcheklein:
@frontexpert static/uploads/images/2017-08-23/post-image.png look too large - compress, optimize? it's way bigger then everything else

plugins/resize-image-plugin/index.js
frontexpert: In merge don't merge the next condition for resizing the image in the build! I made the adjustment

frontexpert:
@shcheklein let me know your feedback.

plugins/resize-image-plugin/index.js
```diff
@@ -77,6 +78,11 @@ module.exports = ({ markdownAST }) => {
...
+        .pop()
```
shcheklein: are we effectively removing it this way?

plugins/resize-image-plugin/index.js
```diff
@@ -98,6 +104,15 @@ module.exports = ({ markdownAST }) => {
         }
       }
 
+      if (imageMaxWidth * 2 > originalSize) {
```
shcheklein: probably should happen only if resize is not specified. Also the code that replaces can be unified with the if below (just calculate the value it is being replaced with ifs).

plugins/resize-image-plugin/index.js
```diff
@@ -67,6 +67,7 @@ module.exports = ({ markdownAST }) => {
           ...
     */
     wrapperImageList.forEach(wrapperImage => {
+      const source = select(`picture > source`, wrapperImage);
```
shcheklein: I see that we have multiple sources above ... why is that? do they have different properties?

src/components/hero/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  srcSet: string;
+  sizes: string;
+  base64?: string;
+  tracedSVG?: string;
+  srcWebp?: string;
+  srcSetWebp?: string;
+  presentationWidth?: number;
+}
+
+interface IGatsbyImageProps {
```
shcheklein: don't like duplicating these interfaces here ... where did you take them? can we export?

src/components/hero/index.tsx
```diff
@@ -0,0 +1,81 @@
+import Image from 'gatsby-image';
```
shcheklein: @iAdramelk could you please review this?

plugins/resize-image-plugin/index.js
frontexpert: The gatsby Remark Image to generate two sources tag one with extension .webp and other with the extension of the image! 
I need this to get the original size image.

plugins/resize-image-plugin/index.js
frontexpert: Ok. I made this

plugins/resize-image-plugin/index.js
frontexpert: I need this to get the original size image value

src/components/hero/index.tsx
frontexpert: I had to declare the types here because by default the fluidObject of Gatsby does not come with the typo of the presentationWidth

src/components/hero/index.tsx
frontexpert: I had to declare the types here because by default the fluidObject of Gatsby does not come with the typo of the presentationWidth

plugins/resize-image-plugin/index.js
shcheklein: do both of them have srcSet?

plugins/resize-image-plugin/index.js
shcheklein: yep, I got that. But it looks like this has a side effects - it removes it from the source.properties ... even if I don't serialize source back down, I would try to avoid this

src/components/hero/index.tsx
shcheklein: not sure I understand this still ... cab you elaborate? can we inherit and include missing fields or something?

plugins/resize-image-plugin/index.js
frontexpert: Yes! But I fixed now to select only the first `picture > source:first-child`

shcheklein:
@frontexpert also, check this https://deploy-preview-103--dvc-blog.netlify.com/r-code-and-reproducible-model-development-with-dvc - looks like hero image background is broken

frontexpert:
> @frontexpert also, check this https://deploy-preview-103--dvc-blog.netlify.com/r-code-and-reproducible-model-development-with-dvc - looks like hero image background is broken

I'm sorry this image I replace with more size! but I switch background transparent by white! I'll replace again with the background  correct

frontexpert:
@shcheklein I replace the image with better width and original background-color

shcheklein:
@frontexpert something is wrong with images sharpness and sizes:

one is from https://blog.dataversioncontrol.com/r-code-and-reproducible-model-development-with-dvc-1507a0e3687b

second is from https://deploy-preview-103--dvc-blog.netlify.com/r-code-and-reproducible-model-development-with-dvc

Screen Shot 2020-02-11 at 8 45 54 AM
Screen Shot 2020-02-11 at 8 45 59 AM

shcheklein:
@frontexpert regarding the last comment - size is probably fine and expected, but sharpness is something I would check

frontexpert:
> @frontexpert something is wrong with images sharpness and sizes:
> 
> one is from https://blog.dataversioncontrol.com/r-code-and-reproducible-model-development-with-dvc-1507a0e3687b
> 
> second is from https://deploy-preview-103--dvc-blog.netlify.com/r-code-and-reproducible-model-development-with-dvc
> 
> Screen Shot 2020-02-11 at 8 45 54 AM
> 
> Screen Shot 2020-02-11 at 8 45 59 AM

It is OK! Because de second have not width enough and  render 1/2px of the original width

shcheklein:
@frontexpert the problem that even though we do 1/2 of the size, it looks even more blurry to me. Either I'm missing something (most likely) about the way images are being processed, or this is Gatsby does not expect intrusions from the plugin, or something else.

@fabiosantoscode could you review this? any insights?

frontexpert:
@shcheklein We are not manipulating the image in the process, but after the process!

frontexpert:
I think we just render original size image when it is smaller than max-width default is enough this issue

What do you think @shcheklein and @fabiosantoscode?

fabiosantoscode:
This looks good to me. I think we should centralise our types somewhere, but that's certainly out of scope.

shcheklein:
@frontexpert 

this is how it should like when downsized properly:

Screen Shot 2020-02-11 at 9 57 19 AM

shcheklein:
something is wrong with the way we downsize images, it might be even something wrong with `=size` support because of this.

shcheklein:
@frontexpert @fabiosantoscode 

Screen Shot 2020-02-11 at 10 01 31 AM

frontexpert:
I will review all base code

frontexpert:
The blurry is because this image is very small. The original size image is 684px and when we downsize the image the pixels don't have resolution enough for not blur

Do you can see the image below has size 1200px and is render with 200px  

image

shcheklein:
@frontexpert my comment https://github.com/iterative/blog/pull/103#issuecomment-584770288 compares two images of the same size (1/2 of the origin). To the right is made in Sketch. Screenshot scale is 100%, not surprises.

shcheklein:
@frontexpert so, to reiterate - image to the right is _rendered_ at 380px and it behave as I would expect it to.

frontexpert:
> @frontexpert my comment [#103 (comment)](https://github.com/iterative/blog/pull/103#issuecomment-584770288) compares two images of the same size (1/2 of the origin). To the right is made in Sketch. Screenshot scale is 100%, not surprises.

I think that the render of Sketch is different from the navigator!

shcheklein:
> I think that the render of Sketch is different from the navigator!

could you elaborate please? I'm not sure I got the point.

frontexpert:
> > I think that the render of Sketch is different from the navigator!
> 
> could you elaborate please? I'm not sure I got the point

The sketch will be an image manipulation program when you reduce it, it already resets the pixels to improve the image sharpness!
And browsers are unable to redefine image pixels to improve sharpness

shcheklein:
> And browsers are unable to redefine image pixels to improve sharpness

both of them scale image down, both of them can do absolutely the same with pixels. Am I missing something?

frontexpert:
The same image downsize 50% on the Medium

image

image

shcheklein:
@frontexpert works fine for me. It's not blurry:

Screen Shot 2020-02-11 at 1 48 19 PM

frontexpert:
This is rendering better on firefox browser

image

frontexpert:
this is preview PR on Firefox, do you can to test in another browser 

image

shcheklein:
Looks like it was cache or something else, not sure. It renders them fine now.

src/components/post/index.tsx
```diff
@@ -86,19 +86,9 @@ function Post({ html, timeToRead, frontmatter, fields }: IBlogPostData) {
...
-              fluid={picture.childImageSharp.fluid}
-              className={styles.picture}
-            />
-          </div>
-          {pictureComment && (
-            <div className={styles.pictureComment}>{pictureComment}</div>
-          )}
-        </>
-      )}
+
```
iAdramelk: in the old version, if there was no picture, this block was not rendered, in the new version this check is removed. Is this a bug or a feature?

frontexpert: I added this condition again!

src/components/hero/index.tsx
frontexpert: I moved the interfaces to template blog

frontexpert:
> Looks like it was cache or something else, not sure. It renders them fine now.

Good

frontexpert:
@shcheklein Can you assign me the next issues? just feeling a lack of tasks now.
Thanks.

shcheklein:
@frontexpert looks good to me. The last thing - can we add more tests like @fabiosantoscode did for the plugin?

frontexpert:
> @frontexpert looks good to me. The last thing - can we add more tests like @fabiosantoscode did for the plugin?

The tests of the plugin are testing the regex and everyone is already covered! 
What other tests do you suggest

shcheklein:
@frontexpert okay, let's merge it then!

frontexpert:
which issues/tickets can I work on next?

PR #112, opened 2020-02-13
pagination

### LINKED ISSUE #110: add pagination to all pages
As we post more and more stuff we need to enable some pagination.

fabiosantoscode:
This adds pagination to `/` and tag pages. The URLs are:

`/`, `/page/2`, `/page/3`, ...

And

`/tags/xyz`, `/tags/xyz/page/2`, ...

To keep things nice and static, we pass `skip` and `limit` context properties for the tag and home pages, and create them in a loop instead of just one.

The PageInfo graphQL type is used as the source of truth for page counts, both when creating the pages and in the paginator itself.

To keep this one simple I tried to make use of GraphQL fragments, and moved types to the component files pertaining to them. I should've probably done this in a separate PR.

In true _me_ fashion, the paginator is ugly and uncentered. I'll have a second pass tomorrow morning, and would appreciate pointers. :zzz: 

Didn't squash the commits to help the reviewing process.

Fixes #110


fabiosantoscode:
This time, the netlify deploy is running at the same time as the tests.

fabiosantoscode:
I just noticed that since I set 20 posts per page in my last commit, we can't actually see the paginator at all. I'm going to set the limit to 5, so that it's at least visible.

iAdramelk:
@fabiosantoscode looks good for me! We need to update design and then we can merge it.

fabiosantoscode:
Ok, things are centered now. On mobile, the links go in a column:

Screenshot from 2020-02-13 16-12-31

Screenshot from 2020-02-13 16-12-36

While on desktop the links go in a row and are separated by 50px, like the posts in the feed itself:

Screenshot from 2020-02-13 16-12-47

Screenshot from 2020-02-13 16-12-43

shcheklein:
@fabiosantoscode good stuff!!

Few comments, mostly UX related:

- first page has a different layout than the 2nd, 3d, etc. We need to take that into account. It mostly means that it's better to limit page not by number of items but number of rows from the user perspective - like each page is 3 full rows.
- Previous/Next - let's rename (at least on the first page) - make it `More items ->` or `Older posts ->` or something like that.
- On the mobile screen - support left and right swipe? Previous and next - let's put them in the corners with some < > icons or something?
- It feels that page reloads itself on the navigation. Is it inevitable? Can we make it smooth that it does reload and jump too much? (cc @iAdramelk - any ideas? what do you think about this?)
- Colors/design - check the `Learn more ->` on the main page.

fabiosantoscode:
I'll see to those points tomorrow. I wouldn't do swipe actions though. On safari for IOS, a swipe is the back button. I don't know how we can allow for both actions with one gesture.

For transitions, it might be possible to change the page live and do a scrollTo(x, y, { behaviour: 'smooth' }). It doesn't need libraries and works great.

fabiosantoscode:
This one was tricky, but on the home page, to compensate for the hero item, there is one less item on the page.

Sadly, when you view the home page with a tablet, you'll see a hole in the last row. This will always happen, because we would need to find a number which is a multiple of both 2 and 3, and which, when 3 is taken out of it, is still a multiple of 2.

Tag pages are fine in this regard since they don't have a hero item.

shcheklein:
@fabiosantoscode looks great!

Let's increase the limit to show one more row, what do you think?

Also, could please check the sitemap generator and rss feed generators, so that they are not affected.

fabiosantoscode:
One more row would look nice.

RSS and sitemap.xml are unaffected.

fabiosantoscode:
9 posts per page, and squashed the commits.

src/constants.js
```diff
@@ -1,4 +1,9 @@
 module.exports = {
   imageMaxWidth: 700,
-  imageMaxWidthHero: 850
+  imageMaxWidthHero: 850,
+
+  // Changing this has a negative SEO impact
+  postsPerPage: 9,
+  // This is so that we know how many posts to remove from a feed page with a hero item on top to make up for its height. MUST be smaller than postsPerPage.
```
shcheklein: comment is too long ... it should not be actually passing the style check?

shcheklein: I mean line itself it too long ... it should be wrapped at 80

src/templates/home-page.tsx
```diff
@@ -0,0 +1,52 @@
+import { graphql } from 'gatsby';
+import React from 'react';
+
+import Layout from '../components/layout';
+import { IPageInfo } from '../components/paginator';
+import Feed, { IFeedPostList } from '../components/feed'; // tslint:disable-line: ordered-imports
```
shcheklein: same here - look like > 80?

gatsby-node.js
```diff
@@ -1,21 +1,26 @@
+/* tslint:disable object-literal-sort-keys */
```
shcheklein: why do we disable it for the whole file?

gatsby-node.js
```diff
@@ -53,22 +70,77 @@ exports.createPages = async ({ graphql, actions }) => {
...
         previous,
+        currentPage: index + 1,
         slug: post.node.fields.slug
       },
       path: post.node.fields.slug
     });
   });
 
-  const tags = result.data.tags.group;
+  function* iterPages({ itemCount, hasHeroItem = false, basePath }) {
```
shcheklein: it feels wrong to include everything into the global gatsy-config - we should probably move this somehwere

fabiosantoscode:
Addressed comments. Looks like none of our lints care about super-long lines, just checked. Opening an issue.

src/utils/iter-pages.js
```diff
@@ -0,0 +1,37 @@
+const { postsPerPage, postsPerRow } = require('./src/constants');
```
shcheklein: what is the regular convention for this? it does not look like an util code to me, or at least not a global/general util ... it's not a component either ... any other place people put this code to?

src/utils/paginated-url.js
```diff
@@ -0,0 +1,11 @@
+// Returns a URL including the page number, given a base path.
```
shcheklein: can we also move this into the paginator directory? does not look like a global util to me as well - it's used in a single very specific place

fabiosantoscode: This is used in gatsby-node as well as the paginator. They both need to compute page URLs.

src/utils/iter-pages.js
fabiosantoscode: I didn't add an extra root folder because it didn't feel like it would warrant that. Placed it here because tag-to-slug is here too.

Another contender is `config/`, which already has some gatsby-specific helper code. I'll place it there.

src/utils/paginated-url.js
shcheklein: kk, up to you ... can gatsby-node export something like `paginator/page-url.js`? or `paginator/util.js`?

fabiosantoscode: I guess it can import `src/components/paginator/page-url.js` as long as it's using ES5 exports.

shcheklein: yep, I would move it to the paginator since it does not look like a global util in any way to me.

shcheklein:
@fabiosantoscode looks like deployment failed ^^

fabiosantoscode:
My bad! Added another requirement for the lint - making sure the requires are resolvable.

src/components/paginator/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    dangerouslySetInnerHTML={{
+      __html: `html { scroll-behavior: smooth; }`
+    }}
+  />
+);
+
+// Turn the arrow around
+const reverse = {
+  transform: 'rotate(180deg)'
+};
```
fabiosantoscode: Ops, should put this in the CSS

fabiosantoscode:
I'm going to add a couple of improvements here.

src/components/paginator/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+          to={paginatedUrl(basePath, currentPage - 1)}
+        >
+          <ArrowSVG style={reverse} />
+          <span>Newer posts</span>
+        </Link>
+      )}
+      {hasNext && (
+        <Link
+          className={cn(styles.link, styles.linkNext)}
+          to={paginatedUrl(basePath, currentPage + 1)}
```
fabiosantoscode: I'm moving this to page creation time, making this component a little bit dumber.

config/iter-pages.js
```diff
@@ -0,0 +1,38 @@
+const { postsPerPage, postsPerRow } = require('../src/constants');
```
shcheklein: utils, config - are super confusing places for this :(

sorry for the nitpicking here, Fabio ... just trying to come up with a better location ... I would put all this stuff into `paginator/utils` ... close to what it's actually doing

fabiosantoscode: No problem! I'll place it there.

fabiosantoscode:
Experiencing some undesired smooth scrolling when clicking back after reaching an article.

shcheklein:
@fabiosantoscode can we consider this is done?

fabiosantoscode:
@shcheklein when you click back from an article you experience smooth scroll. This is not ideal, so I'm working around this.

fabiosantoscode:
There we go. Ready for primetime :)

shcheklein:
Great stuff, @fabiosantoscode !

PR #113, opened 2020-02-13
subscribe: send event when user subscribes

### LINKED ISSUE #38: subscribe to the email list consider using captcha, add GA event
We are getting bots subscribed to the list. Not clear where do they come from yet - but we need to add additional safeguards + add GA properly to see if subscriptions come from blog (and how many).

fabiosantoscode:
This sends a google analytics event when the user subscribes. It's probably easier to test in production.

Closes #38 (which also needs some mailchimp access)

### DIFF
--- a/src/components/subscribe/form/index.tsx
+++ b/src/components/subscribe/form/index.tsx
@@ -1,6 +1,28 @@
-import React from 'react';
+import React, { FormEvent } from 'react';
 
 import styles from './styles.module.css';
+
+const honeyPot = 'b_a08bf93caae4063c4e6a351f6_24c0ecc49a';
+
+interface IFormElements {
+  [honeyPot]: HTMLFormElement;
+}
+
+const sendGAEvent = ({ target: form }: FormEvent<HTMLFormElement>) => {
+  const formElements: IFormElements = (form as any).elements;
+  if (formElements[honeyPot].value) {
+    // It's a bot.
+    return;
+  }
+
+  if (typeof window.ga === 'function') {
+    window.ga('send', 'event', {
+      eventCategory: 'blog-home',
+      eventAction: 'subscribe',
+      transport: 'beacon'
+    });
+  }
+};
 
 export default function SubscribeForm() {
   return (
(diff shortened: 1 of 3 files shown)


shcheklein:
@iAdramelk could check please

@fabiosantoscode does it use the same naming/schema conventions for events as we have already?

fabiosantoscode:
@shcheklein sorry, I didn't check for GA events in the dvc.org repo. It should be up to the existing conventions now.

src/components/subscribe/form/index.tsx
```diff
@@ -1,7 +1,31 @@
...
+}
+
+const sendGAEvent = ({ target: form }: FormEvent<HTMLFormElement>) => {
+  const formElements: IFormElements = (form as any).elements;
+  if (formElements[honeyPot].value) {
+    // It's a bot.
+    return;
+  }
+
+  const ga: (...args: any) => any = (window as any).ga;
```
iAdramelk: We should probably use existing Google Analytics typings from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/google.analytics/index.d.ts instead of using `any`.

fabiosantoscode: A lot better than fighting the type system :)

src/components/subscribe/form/index.tsx
```diff
@@ -1,7 +1,29 @@
...
 
 import styles from './styles.module.css';
 
+const honeyPot = 'b_a08bf93caae4063c4e6a351f6_24c0ecc49a';
+
+interface IFormElements {
+  [honeyPot]: HTMLFormElement;
+}
+
+const sendGAEvent = ({ target: form }: FormEvent<HTMLFormElement>) => {
```
shcheklein: can we generalize the sender into an utils function - will need this in other places. looks like we don't track anything except page views - mind creating a ticket for this?

the first candidate would be to log next/previous page views, social share buttons, comments button click, tags click, etc

fabiosantoscode: Sure!

fabiosantoscode:
Comments addressed and commits squashed

PR #115, opened 2020-02-14
ci: attempt to speed up deployment through the use of cache.

### LINKED ISSUE #109: speedup deployment
It takes ~7 minutes to deploy to Netlify.

It should be done within seconds or better.

@fabiosantoscode @iAdramelk any thoughts?

fabiosantoscode:
Might close #109


fabiosantoscode:
This doesn't work at all.

PR #120, opened 2020-02-18
Frontmatter markup to html #39

### LINKED ISSUE #39: support basic markup in meta fields

cosmn:
After long research into gatsby data pipeline and countless plugins walk-through, I managed to pin-point the best way to hook my markdown to html conversion.

https://www.gatsbyjs.org/docs/node-apis/#onCreateNode

Fixes #39 

Questions and opinions are welcomed.


cosmn:
The default HTML conversion by `remark-html` surrounds everything in a `<p>` tag, with no option to  omit that, which introduces extra styling ( browser imposed margin ).
Should we manually trim the `<p>...</p>` out of the resulting values ?

.gitignore
```diff
@@ -6,3 +6,4 @@ tmp/*
 .env
 .DS_Store
 .idea
+.vscode
```
fabiosantoscode: Lifehack: if you want to ignore things without pushing to a remote .gitignore, you can add them to `.git/info/exclude` with the same syntax. This avoids having a bunch of unrelated things in the .gitignore.

gatsby-node.js
```diff
@@ -86,6 +91,24 @@ exports.onCreateNode = ({ node, actions, getNode }) => {
...
+      node.frontmatter.descriptionLong = markdownToHtml(
+        descriptionLong
+      ).contents;
+    }
+
+    if (descriptionLong) {
+      node.frontmatter.pictureComment = markdownToHtml(
+        descriptionLong
+      ).contents;
+    }
```
fabiosantoscode: I'm confused. Did you mean `if (pictureComment)` and `markdownToHtml(pictureComment)`?

fabiosantoscode:
@cosmn 

> The default HTML conversion by remark-html surrounds everything in a <p> tag

This should be fine, especially if the description has multiple paragraphs. We just need to style the `<p>` tag, which you can do in `.description p {}` in the CSS easily.

Nice work so far!

.gitignore
cosmn: Good to know that. Thank you.
The same thing can be said about the `.DS_Store`, yet its there, unrelated to my OS or editor of choice. Question is: what makes a value popular enough to end up in `.gitignore` and not in a personal `.git/info/exclude`

gatsby-node.js
cosmn: Of course. copy/paste/replace gone wrong here
Thank you for catching that

cosmn:
@fabiosantoscode There's two ways going about it. Either remove the `<p>` because it wasn't there to begin with and each field has its own wrapper already, or leave it there and CSS reset it, which seems like an unnecessary band aid.

Thank you for your input. Much appreciated.

.gitignore
fabiosantoscode: I don't know. I do think we should move .idea and .DS_Store out, but opinions will vary. .DS_Store is pretty widespread in .gitignore files I've seen, since they're always placed there when you use Mac OS.

fabiosantoscode:
@cosmn

Given that there might be multiple paragraphs, more care should probably be taken there.

```
.foo p {
  margin: 0;
  & + p {
    margin-top: [default margin];
  }
}
```

cosmn:
@fabiosantoscode you're right.
I just checked, and we get multiple paragraphs when the text contains `\r\r`, which makes sense.
Your css snippet will do the job. 👍

iAdramelk:
@cosmn great work!

Small comment from me. We use description, descriptionLong and other frontmatter fields as part of our `og:` tags values in the header. If we place html there it will break them. So we probably need to create separate text and html version of these fields. Like:

- `description` - replace version with markdown with ∏lain text version.
- `descriptionHTML` - html version.

iAdramelk:
Also, right now remarkHTML didn't use our remark plugins from `gatsby-config.js`, but it is probably not a big problem.

fabiosantoscode:
Great point!

This might fit the bill, maybe we should test it: https://github.com/teramotodaiki/remark-plain-text

.gitignore
shcheklein: I would vote to keep them - do not see any good reason to not include to be honest.

@cosmn just put a new line please and `.vscode`

cosmn:
@iAdramelk what I found is that `descriptionLong` is only used within `<body>`. Could you point to to a location that confirms otherwise ? if not, renaming is not required, unless it serves well in other cases.
Regarding remarkHTML, the last thing I wanted is to introduce another node module, but I figured it would be of use since we're relying on `remark` and we didn't had a a way to convert to html.
Another option would be to use `mdast-util-to-hast` then `hast-util-to-html`, both modules being already installed and used by `gatsby-transformer-remark`

@fabiosantoscode Honestly I would rather have multiple fields each with its own purpose than to convert from markdown back to plain. This one is debatable.

Thank you all for your input.

iAdramelk:
@cosmn I may remember incorrectly about `descriptionLong`, but `description` and `title` are used here for example https://github.com/iterative/blog/blob/master/src/templates/blog-post.tsx#L80-L81

cosmn:
@iAdramelk exactly. but not `descriptionLong`. The only frontmatter we're converting is `descriptionLong` and `pictureComment`, which are only used within `<body>`

.gitignore
cosmn: @shcheklein is there something wrong with the way it is now ?

shcheklein: it's a minor thingy ... it's a good practice to have a new line at the of the file, even Github highlights it:

Screen Shot 2020-02-18 at 2 34 45 PM

shcheklein:
@cosmn good stuff! as an example for this feature can we pick at least one post (check the  blog.dataversioncontrol.com archive) to pick the one with a link in the picture comments, for example) and apply some markup, please? it's hard to review w/o seeing what it generates.

.gitignore
cosmn: understood. Done.

content/blog/2019-08-01-july-19-dvc-heartbeat.md
```diff
@@ -8,13 +8,16 @@ description: |
 descriptionLong: |
   Every month we are sharing here our news, findings, interesting reads,
   community takeaways, and everything along the way.
-  Some of those are related to our brainchild DVC and its journey. The others
+
+  Some of those are related to our brainchild [DVC](https://dvc.org) and its journey. The others
```
shcheklein: looks good as a test - let's remove the link before we merge :)

cosmn:
@shcheklein I've updated [july-19-dvc-heartbeat](http://localhost:8000/july-19-dvc-heartbeat) to match its [old blog equivalent](https://blog.dataversioncontrol.com/july-19-dvc-%EF%B8%8Fheartbeat-b178bf8454e7) in order to have a testable page.

One thing to keep in mind is that the markdown syntax may interfere with the yaml syntax.
Example:
```yaml
pictureComment: [a link](/page)
```
would yield a processing error, which can be avoided by re-wording it
```yaml
pictureComment: Here's [a link](/page)
```
or a pipe
```yaml
pictureComment: |
    [a link](/page)
```

content/blog/2019-08-01-july-19-dvc-heartbeat.md
cosmn: it's not a test, this `descriptionLong` and `pictureComment` is updated according to its [old post equivalent](https://blog.dataversioncontrol.com/july-19-dvc-%EF%B8%8Fheartbeat-b178bf8454e7). Do I still remove it ?

shcheklein: good, let's keep it then! thanks :)

shcheklein:
@fabiosantoscode it seems you need to update RSS feed serializer (`gatsby-plugin-feed` options). Looks like there is a bug in it as well ` edge.node.descriptionLong || edge.node.descriptionLong` that should be updated.

shcheklein:
thanks @cosmn !

PR #122, opened 2020-02-19
draft: custom express server

fabiosantoscode:
This is a draft just to show what custom API endpoints look like in a local express server. Discuss in https://github.com/iterative/dvc.org/issues/992


api/index.js
```diff
@@ -0,0 +1,11 @@
+const express = require('express');
+
+const api = express.Router();
+
+api.get('/api/example-function', (req, res) => {
+  res.json({
+    amAPI: true
+  });
+});
```
fabiosantoscode: We can remove the `/api` if we want by mounting this in `/api`.

We can also add an error handler just for these here routes, which can talk to a crash reporter and respond with JSON.

Also we can mount a bunch of middleware on this Router. Like adding cache headers, a memory/redis cache, providing etags for calling the github API, creating a unique request id for traceability, etc.

server.js
```diff
@@ -0,0 +1,16 @@
...
+const gatsyExpress = require('gatsby-plugin-express');
+const app = express();
+
+const api = require('./api');
+
+app.use(
+  express.static('public/', {
+    trailing: false // Don't do a pesky trailing slash redirect
+  })
+);
```
fabiosantoscode: We could use `gatsby-plugin-express` here too. Sadly it's only for prod. I thought it worked in development mode as well.

gatsby-config.js
```diff
@@ -170,6 +171,9 @@ if (process.env.CONTEXT === 'production') {
 }
 
 module.exports = {
+  developMiddleware: app => {
+    app.use(api);
+  },
```
fabiosantoscode: This is where the magic happens. Since the API is a Router, it's like a middleware, which we can mount it onto gatsby in development, something I didn't know was possible before.

api/index.js
fabiosantoscode: And obviously we can split this into several files and test them individually, where api/index gathers functions and mounts them into the router.

shcheklein:
@iAdramelk @jorgeorpinel any thoughts on this setup?

iAdramelk:
@shcheklein I still prefer full static approach, but this looks ok too.

jorgeorpinel:
and I still think we can just separate the web app from the api. Separate apps = less complexity, custom server, conditional logic, decoupled load capacity, etc. But this works for sure, thanks for the demo Fabio.

shcheklein:
@jorgeorpinel 

> Separate apps = less complexity

you can see it from a different angle. For example, it means that you need to deploy two things separately (including managing potential discrepancy in versions), it breaks your local workflow (or at least complicates it significantly)

jorgeorpinel:
I see decoupling as an improvement. No risk of webapp bug breaking api for example, and handling loads also, if api gets overloaded, webapp doesn't have to slow down. Deploying 2 separate apps is not a problem, CI/CD linked to each repo takes care of it. Locally you only need to run one or the other. local dev webapp can connect to prod api and vice versa.

jorgeorpinel:
If they used the same patterns and technologies I wouldn't mind that its the same deploment/monorepo app but since they're just completely different in their implementation and nature, I think it makes sense to just completely split them. They don't even overlap in the current code base anyway, from what I understand.

jorgeorpinel:
> discrepancy in versions

This on is a potential problem, yes. But I doubt we'll have so many api changes that the version will matter that much. The few latest api and few latest web app versions should always be compatible I expect.

jorgeorpinel:
Final thought: they can remain in a single monorepo to solve the version discrepancy risk. Still CI can deploy each app to separate places. There will just be some unnecessary deployments (e.g. when a PR where web app has updates but api doesn't gets merged, still both get (re)deployed.

shcheklein:
> They don't even overlap in the current code base anyway, from what I understand.

they will

> but since they're just completely different in their implementation and nature, I think it makes sense to just completely

yep, you are right

> ... makes sense to just completely split them

makes sense, but'll complicate everything. There are some advantages indeed (like isolation), but in general it's more complicated to deploy into two different environments and running things locally (hitting prod from dev is an anti-pattern,  CI/CD linked to each repo takes care of it - you will need to write this first, preview apps will break for example, or will have to hit different environments).


Not that I'm against this, but would try to find a simpler solution first.

shcheklein:
@jorgeorpinel to clarify, discrepancy in the deployment we'll need to deal with is that one preview app won't know anything (by default) about the second one. It'll keep hitting prod APIs.

jorgeorpinel:
Yeah good points. I guess they both bring different complexities. I'm not sure which one is better haha.

My other Q is, if the API is basically a proxy to 3rd party APIs and only used for the community page and maybe blog posts, why do we need to do HTTP requests within the same API? Why not just import a module that does the fetching/processing of the 3rd party APIs?

shcheklein:
> why do we need to do HTTP requests within the same API? Why not just import a module that does the fetching/processing of the 3rd party APIs?

mostly due to rate limits and response times, meaning we want to cache results aggressively. mostly means we need a server with some server-side caching mechanism

fabiosantoscode:
@jorgeorpinel 

> makes sense to just completely split them

It does make sense in terms of independent scaling, load management etc, however since our website is mostly static this is a bit overkill. The cache will ensure we only really get a few requests per minute.

WRT bugs/crashes in the API affecting the site, I think a few unit tests using nock to mock external APIs would greatly mitigate this. The cyclomatic complexity of the API is small, so there isn't much space for bugs there. And if github itself is down, we can always reply with stale content (CDN might be able to do this with no extra code if we have one) and nobody bats an eye.

For bugs in the site affecting the API, I don't see that happening. The site is a bunch of inert, static files in production, being served by an express file handler, which is as battle tested as you can get. The only thing that might happen is a typo in the URL used to call the API, but I think it's a tad unlikely that someone would change a URL without testing it.

jorgeorpinel:
> we want to cache (API) results aggressively

Could use a memory cache for this (@shcheklein).

> WRT bugs/crashes... there isn't much space for bugs... 
> bugs in the site affecting the API, I don't see that happening...

Fabio, I agree with you in general, but I don't think it's safe to assume there will be no bugs because we don't see them now. You typically don't see bus coming 🙂

server.js
fabiosantoscode: Even better: `serve-handler`. It removes trailing slashes with a redirect instead of passively leaving them there.

fabiosantoscode:
Closing this since we have an express server now.

PR #129, opened 2020-02-21
escape HTML special characters and tags within external link attributes

fabiosantoscode:
I was trying to remove the `<p><section>` sequence in this plugin, since it's invalid and forces the browser to change the HTML.

It was taking too long so I left this change behind: attributes in the external-link tag weren't being escaped before being inserted into the tree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "@reach/tooltip": "^0.5.4",
     "@types/google.analytics": "^0.0.40",
     "classnames": "^2.2.6",
+    "escape-html": "^1.0.3",
     "gatsby": "^2.19.19",
     "gatsby-image": "^2.2.39",
     "gatsby-plugin-twitter": "^2.1.18",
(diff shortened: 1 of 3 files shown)


shcheklein:
@fabiosantoscode please, merge the master/rebase

also, could you please elaborate a bit. I got the idea behind escaping - good catch. Not sure about the p/section issue and how did you find this problem in the first place?

iAdramelk:
Looks good, but maybe little too paranoid. I don't think that we will add md-files with inlined js in repo, and if third party can accept PRs without checks from our side, then `external-link` is not out main problem.

@shcheklein from my understanding `p/section` will generate warnings to the console, like tooltips doing now.

fabiosantoscode:
Escaping is not always about being paranoid :) In this case, it's for when we want to add `<` and `>` to those descriptions without issues.

iAdramelk:
@fabiosantoscode fair point.

shcheklein:
@fabiosantoscode please, merge/rebase. Let's create a ticket for the `p/section` issue. I wonder if we can replace `section` with something else?

PR #130, opened 2020-02-21
Fix #128 Inifinite rebuilds

### LINKED ISSUE #128: changes in post's metadata break hot reload and dev mode
try `content/blog/2019-08-01-july-19-dvc-heartbeat.md` file edits to the `pictureComment` can easily get rebuild into infinite cycle `running queries`.

no information why is it happening, no errors - super annoying and is breaking editors workflow. 

@iAdramelk  do you know if there is an easy way to avoid this in this future, easy way to run some tests for this? (this is not the first time it happens)?

iAdramelk:
Updating `gatsbyjs` to the latest version fixed this bug for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@reach/tooltip": "^0.5.4",
     "@types/google.analytics": "^0.0.40",
     "classnames": "^2.2.6",
-    "gatsby": "^2.19.1",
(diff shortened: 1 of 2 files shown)


shcheklein:
thanks @iAdramelk !

PR #135, opened 2020-02-28
hosting: use heroku hosting

fabiosantoscode:
(Also changed the a type in the post because typescript started complaining)

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,10 +1,17 @@
 {
+  "addons": [],
   "buildpacks": [
     {
       "url": "heroku/nodejs"
-    },
-    {
-      "url": "https://github.com/heroku/heroku-buildpack-static"
     }
-  ]
+  ],
+  "env": {},
+  "formation": {
+    "web": {
+      "quantity": 1
+    }
+  },
+  "name": "dvc-blog-production",
+  "scripts": {},
+  "stack": "heroku-18"
 }
--- a/package.json
+++ b/package.json
@@ -4,11 +4,13 @@
   "description": "Blog for dvc.org",
   "version": "0.1.0",
   "license": "Apache-2.0",
+  "engines": {
+    "node": ">=12.0.0"
+  },
   "scripts": {
     "build": "gatsby build",
     "develop": "gatsby develop",
-    "serve": "gatsby serve",
-    "start": "yarn develop",
+    "start": "node ./src/server/server.js",
     "typecheck": "tsc --noEmit --skipLibCheck",
     "test": "jest",
     "lint-ts": "tslint --project .",
@@ -42,6 +44,7 @@
     "@types/google.analytics": "^0.0.40",
     "classnames": "^2.2.6",
     "escape-html": "^1.0.3",
+    "express": "^4.17.1",
     "gatsby": "^2.19.19",
     "gatsby-image": "^2.2.39",
     "gatsby-plugin-twitter": "^2.1.18",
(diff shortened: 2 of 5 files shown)


fabiosantoscode:
Builds are failing because of the node engine (I changed to 12+ for performance, and longer term maintenance).

For the builds back at my fork have a look at:

 - tests: https://dashboard.heroku.com/pipelines/870ffb13-217c-4c46-9846-7db57fe65722/tests/22
 - build (slower because I changed package.json) https://dashboard.heroku.com/apps/dvc-blog-production/activity/builds/92f44987-8e7e-46aa-b43b-1ddf47077d9b

shcheklein:
@iAdramelk can you take a look please?

@fabiosantoscode what is the plan on merging this? we will to update CloudFlare simultaneously, right? We probably need to update CloudFlare's SSL cert to Heroku. Let's discuss the merge plan in Slack.

iAdramelk:
@fabiosantoscode @shcheklein  It works, but why do we need express for the static blog? Is it only for trailing slashes? Did we try to test slashes without express? now.sh didn't have slashes and it serves static, I'm pretty sure that slashes is not the gatsby problem, but the Netlify one.

fabiosantoscode:
The static handler we're using here is from Zeit and it's what they use in now.sh to serve statics.

Netlify wasn't doing anything about slashes. If we kept npm start the same (Gatsby develop) it would force slashes upon the URLs.

I could've used the static handler without the express server but it wouldn't do redirects. Which reminds me that I didn't add redirects.

iAdramelk:
@fabiosantoscode maybe we can use static buildback and do redirects with it?  https://github.com/heroku/heroku-buildpack-static

Sorry to be picky, but I still think that custom express servers with static websites are generally bad idea. We can use it if we don't have other options, but if it possible to do something without it, then I think it's better to do it without it.

fabiosantoscode:
I think you're right to be skeptical here. But much of this came from our discussion over at the dvc.org repository, where we're going to need:

 - redirects from subdomains like man.dvc.org
 - some API endpoints

We can achieve the latter using two buildpacks (but IIRC this means two dynos and therefore more cost). The former, we can't do unless we create a buildpack.

Also remember, that we are comfortably behind cloudflare, which normalises requests (improving security) and caches things (improving requests per second).

I've been serving things using node for a long time and no issues have come about. It has very good performance and security properties especially behind a reverse proxy. Serve in particular is very battle tested

iAdramelk:
@fabiosantoscode I see you point, and after we merge blog and main site it will not be an issue (if we don't find a way to do redirects and api endpoint another way). But for now I think it is an unnescessary complication. So if we can do it without express, I'd prefer to do it this way. The more we became dependent on Express, the harder it will become to remove it later.

P. S. Con you please check `heroku-static` redirects format above, maybe we can use them to solve problem with redirects for subdomains that we can't do on `now.sh`?

fabiosantoscode:
If we could run a "cloudflare emulator" which would give us workers in development and in PR environments (maybe something like [this](https://www.npmjs.com/package/@lifeisaloha/cloudworker)?), we could easily do this.

WRT heroku-static:

I've double-checked their docs and their source code.

Have a look here to see the syntax doesn't have a host in it:
https://github.com/heroku/heroku-buildpack-static#custom-redirects

I've also looked at their nginx configuration, and they aren't doing this at all as far as I can tell.

shcheklein:
> The more we became dependent on Express, the harder it will become to remove it later.

not sure I understand the problem depending on it. Could you elaborate please a little bit?

PR #137, opened 2020-02-28
server: use cache headers

fabiosantoscode:
This sets cache headers so cloudflare can cache the HTML and assets of the site.

### DIFF
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -19,6 +19,17 @@ const app = express();
 const serveOptions = {
   directoryListing: false,
   etag: true,
+  headers: [
+    {
+      headers: [
+        {
+          key: 'Cache-Control',
+          value: 'public, max-age=3600'
+        }
+      ],
+      source: '*'
+    }
+  ],
   public: path.join(__dirname, '../../public'),
   trailingSlash: false
 };


fabiosantoscode:
@shcheklein this didn't get deployed. We need to enable Heroku's CI for this repository (it was already enabled for my repository, but I can't enable it here since I'm not admin)

shcheklein:
CI check fails on master: 

```
error dvc-blog@0.1.0: The engine "node" is incompatible with this module. Expected version ">=12.0.0". Got "10.16.3"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

also the cache control for * for one hour looks a bit too aggressive (I'm concerned that we might forget to remove it for APIs for example)? What headers does Gatsby itself sets if any? ETAG? (so that browser can ping with HEAD)?

shcheklein:
@fabiosantoscode see the previous comment please.

fabiosantoscode:
@shcheklein 

 - I'm updating Circle's node version right now
 - cache-control header is only for the static files. The API will have its own cache headers. If we forget we will get too many requests to it, so we best not forget it :)
 - Yes one hour might be too aggressive. We have less than 100 posts so we can afford to lower this to about 10 or 15 minutes.
 - gatsby sets a max-age=0 cache header paired with an etag, such that the upstream revalidates everything all the time. I don't think this is very optimal, since revalidation requests are still requests that are coming in.

PR #139, opened 2020-03-03
deployment: force browser cache revalidation and purge cache after deploys

fabiosantoscode:
This enables an instant deploy, where the browser always has the latest version of everything. The trade-off is that it has to always revalidate. It's possible to not have mandatory revalidation on the content-addressible parts (JS/CSS compiled chunks, generated thumbnails), but that's a story for another time I guess.

I've already:

 - Checked that both chrome and firefox are spec-compliant and always revalidate when given these headers
 - Checked that the CLOUDFLARE_TOKEN and CLOUDFLARE_ZONE_ID are secret when set at the heroku app level
 - Added the above variables to the production app, to make this script work (this token only has the capability of purging cache, so if it's leaked not much damage can be done)
 - Checked that the script works by running it myself

### DIFF
--- a/scripts/heroku-postdeploy.sh
+++ b/scripts/heroku-postdeploy.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# This script runs after each heroku deploy. If the deploy is the
+# production heroku app (the only one with the below variables)
+# the cache gets cleared.
+#
+# To clear the cache yourself, you can use the button in the
+# cloudflare dashboard ("Caching tab > Purge everything"), or run
+# this script with the required environment variables:
+#
+# - CLOUDFLARE_TOKEN: a token with the "Zone.Cache Purge" permission.
+# You can generate this token in "My Profile > API Tokens"
+#
+# - CLOUDFLARE_ZONE_ID: The zone ID to purge. You can find it in the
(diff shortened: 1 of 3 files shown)


scripts/heroku-postdeploy.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [ -z "${CLOUDFLARE_TOKEN:-}" ]; then
+  exit 0
+fi
+
+curl --fail \
```
fabiosantoscode: --fail is to make curl exit non-zero when the status code is not in the 200 range. `set -e` above makes this non-zero exit force the script to exit non-zero as well, propagating the error.

scripts/heroku-postdeploy.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
shcheklein: let's put some docs here describing what this script is about, how to setup it, etc -- so that we don't get lost

fabiosantoscode: Got it!

fabiosantoscode: It's documented now.

PR #140, opened 2020-03-04
deployment: clear cache after each deploy

fabiosantoscode:
So, yesterday I found that the postdeploy script runs exactly once after the app is deployed, and not in subsequent deployments of the same app. This makes it useless for the use case of clearing cache after the service starts.

I've scoured the Heroku documentation looking for an alternative, but only found releases, which run just before the deployment. There's also a way to run a separate dyno alongside your app, but it can't terminate after doing its thing. One-off dynos can only be run by executing the `heroku run` command, so no luck there either.

So after a lot of digging, I had to come to terms with the fact that the cache has to be cleared in the dyno.


package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "gatsby build",
     "develop": "gatsby develop",
-    "start": "node ./src/server/server.js",
+    "start": "./scripts/heroku-clear-cache.sh && node ./src/server/server.js",
```
shcheklein: name for the script is confusing ... like we are clearing the Heroku's cache

also, this configuration means that it's make hard to run it locally, especially Windows

fabiosantoscode: I'll translate the script to javascript then.

shcheklein: yep, let's make sure that we skip it when we run locally (and it doesn't fail)

fabiosantoscode: There

package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "gatsby build",
     "develop": "gatsby develop",
-    "start": "node ./src/server/server.js",
+    "start": "./scripts/heroku-clear-cache.sh; node ./src/server/server.js",
```
shcheklein: change the name?

scripts/clear-cloudflare-cache.js
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env node
+
+// This script runs after each heroku deploy. If the deploy is the
```
shcheklein: update?

package.json
fabiosantoscode: My bad!

PR #143, opened 2020-03-06
deployment: deploy statics to S3

fabiosantoscode:
So here's what I've been working on this week.

It's not ready to merge, because the script that destroys the PR isn't working for some reason. I can't debug it since there are no logs in the destroy script, so on monday, I'll set up an ngrok and POST stuff to my laptop so I can read errors.

## Strategy

I've replaced the build command only on heroku, such that it:

1. Downloads the public/ folder from S3 (if it's a cold PR it gets the production folder since there are no secrets there)
2. Runs gatsby build
3. Uploads the public/ folder to S3.

The production server then merely proxies to S3. It doesn't work for your local public/ folder.

## Timings

Production build: 3min
First PR build: 4min
Subsequent builds of the same PR: 3min

Downloading and uploading from S3 takes very little time. I was able to increase the request concurrency to 50, so the initial download only takes 10 seconds. Uploading to an already-hot S3 bucket takes around 3 seconds since most files will already be there.

## Drawbacks

In AWS, I've created two IAM users, one for production (which has write access to the production prefix of my bucket), and another for PRs (which has access to $bucket/pulls/*). This means that if someone can leak the tokens in their build (by changing the build command) they can mess with other people's PRs. Therefore, we shouldn't enable automatic PR builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ tmp/*
 .DS_Store
 .idea
 .vscode
+404.html
--- a/scripts/deploy-with-s3.js
+++ b/scripts/deploy-with-s3.js
@@ -1,0 +1,124 @@
+#!/usr/bin/env node
+'use strict';
+
+/**
+ * Build gatsby site and deploy public/ to s3.
+ *
+ * The S3 path of the deployment depends on the HEROKU_APP_NAME variable,
+ * which is passed to PRs by heroku, but you can set locally too.
+ *
+ * With HEROKU_APP_NAME: /dvc-org-pulls/$HEROKU_APP_NAME
+ * Without HEROKU_APP_NAME: /dvc-org-prod
+ *
+ * Needs following environment variables:
+ *
+ *  - S3_BUCKET: name of the bucket
+ *  - AWS_REGION: region of the bucket
+ *  - AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY: auth token to access the bucket.
+ *  - HEROKU_APP_NAME: (optional) app name to specify the ID of the PR if any.
+ **/
+
+const path = require('path');
+const { execSync } = require('child_process');
+const { remove, move } = require('fs-extra');
+const { s3Prefix, s3Bucket, s3Client } = require('./s3-utils');
+
+const rootDir = path.join(__dirname, '..');
+const cacheDir = path.join(rootDir, '.cache');
+const publicDir = path.join(rootDir, 'public');
+
+function run(command) {
+  execSync(command, {
+    stdio: ['pipe', process.stdout, process.stderr]
+  });
+}
+
+function syncCall(method, ...args) {
+  return new Promise((resolve, reject) => {
+    const synchroniser = s3Client[method](...args);
+    synchroniser.on('error', reject);
+    synchroniser.on('end', resolve);
+  });
+}
+
+async function prefixIsEmpty(prefix) {
+  try {
+    await s3Client.s3
+      .headObject({
+        Bucket: s3Bucket,
+        Prefix: prefix + '/index.html'
+      })
+      .promise();
+    return false;
+  } catch (e) {
+    return true;
+  }
+}
+
+async function downloadFromS3(prefix) {
+  try {
+    console.log(`downloading public/ from s3://${s3Bucket}/${prefix}`);
+    console.time('download from s3');
+    await syncCall('downloadDir', {
+      localDir: publicDir,
+      s3Params: {
+        Bucket: s3Bucket,
+        Prefix: prefix
+      }
+    });
+    console.timeEnd('download from s3');
+  } catch (downloadError) {
+    console.error('Error downloading initial data', downloadError);
+    // Don't propagate. It's just a cache warming step
+  }
+}
+
+async function uploadToS3() {
+  console.log(`Uploading public/ to s3://${s3Bucket}/${s3Prefix}`);
+  console.time('upload to s3');
+  await syncCall('uploadDir', {
+    localDir: publicDir,
+    deleteRemoved: true,
+    s3Params: {
+      Bucket: s3Bucket,
+      Prefix: s3Prefix
+    }
+  });
+  console.timeEnd('upload to s3');
+}
+
+async function main() {
+  const emptyPrefix = await prefixIsEmpty(s3Prefix);
+
+  // First build of a PR is slow because it can't reuse cache.
+  // But we can download from prod to warm cache up.
+  const cacheWarmPrefix = emptyPrefix ? 'dvc-org-prod' : s3Prefix;
+
+  await downloadFromS3(cacheWarmPrefix);
+
+  try {
+    run('yarn build');
+  } catch (buildError) {
+    // Sometimes gatsby build fails because of bad cache.
+    // Clear it and try again.
+
+    console.error('------------------------\n\n');
+    console.error(buildError);
+    console.error('\nAssuming bad cache and retrying:\n');
+
+    await remove(cacheDir);
+    await remove(publicDir);
+    run('yarn build');
+  }
+
+  await move(path.join(publicDir, '404.html'), path.join(rootDir, '404.html'), {
+    overwrite: true
+  });
+  await uploadToS3();
+  await remove(publicDir);
+}
+
+main().catch(e => {
+  console.error(e);
+  process.exit(1);
+});
--- a/src/server/server.js
+++ b/src/server/server.js
@@ -1,45 +1,96 @@
 /* eslint-env node */
 
+'use strict';
+
 /*
- * Production server.
+ * Production server. Proxies to S3 depending on HEROKU_APP_NAME (see
+ * scripts/deploy-with-s3.js)
  *
  * NOTE: This file doesn't go through babel or webpack. Make sure the syntax and
  * sources this file requires are compatible with the current node version you
  * are running.
+ *
+ * Required environment variables:
+ *
+ *  - S3_BUCKET: name of the bucket
+ *  - AWS_REGION: region of the bucket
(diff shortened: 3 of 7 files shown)


scripts/deploy-with-s3.js
```diff
@@ -0,0 +1,107 @@
...
+      Bucket: s3Bucket,
+      Prefix: s3Prefix
+    }
+  });
+  console.timeEnd('upload to s3');
+}
+
+async function main() {
+  const emptyPrefix = await prefixIsEmpty(s3Prefix);
+  // TODO if the prefix is empty, we can have S3 copy everything
```
shcheklein: let's not put is a TODO in code - let's create a ticker for this

scripts/deploy-with-s3.js
```diff
@@ -0,0 +1,107 @@
...
+    run('yarn build');
+  } catch (buildError) {
+    console.error('------------------------\n\n');
+    console.error(buildError);
+    console.error('\nAssuming bad cache and retrying:\n');
+    // Sometimes gatsby build fails because of bad cache.
+    // Clear it and try again.
+    await remove(cacheDir);
+    await remove(publicDir);
+    run('yarn build');
```
shcheklein: minor: is there a lib that wraps a call into retry?

scripts/deploy-with-s3.js
```diff
@@ -0,0 +1,107 @@
...
+  const cacheWarmPrefix = emptyPrefix ? 'prod' : s3Prefix;
+
+  await downloadFromS3(cacheWarmPrefix);
+  try {
+    run('yarn build');
+  } catch (buildError) {
+    console.error('------------------------\n\n');
+    console.error(buildError);
+    console.error('\nAssuming bad cache and retrying:\n');
+    // Sometimes gatsby build fails because of bad cache.
```
shcheklein: minor: probably, move comment up

scripts/destroy-pr.js
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const { s3Prefix, s3Bucket, s3Client } = require('./s3-utils');
+
+async function main() {
+  console.log(`Deleting s3://${s3Bucket}/${s3Prefix}...`);
```
shcheklein: we can actually setup a lifecycle policy to destroy stuff after a month just to be safe ... may be even remove this code

shcheklein:
> This means that if someone can leak the tokens in their build (by changing the build command) they can mess with other people's PRs. Therefore, we shouldn't enable automatic PR builds.

could you elaborate on this please, not sure I got the idea.

app.json
```diff
@@ -12,6 +12,8 @@
     }
   },
   "name": "dvc-blog-production",
-  "scripts": {},
+  "scripts": {
+    "pr-predestroy": "./scripts/destroy-pr.js"
```
shcheklein: see my comment? may be we can simplify with lifecycle policies. It also beneficial since PR tend to expire after 5 days and it's actually better to reuse bucket if we get it up again.

scripts/deploy-with-s3.js
```diff
@@ -0,0 +1,107 @@
...
+      localDir: publicDir,
+      s3Params: {
+        Bucket: s3Bucket,
+        Prefix: prefix
+      }
+    });
+    console.timeEnd('download from s3');
+  } catch (e) {
+    console.error('Error downloading initial data');
+    console.error(e);
```
shcheklein: is it possible to do `console.error(msg, e`)?

scripts/deploy-with-s3.js
```diff
@@ -0,0 +1,107 @@
...
+
+async function main() {
+  const emptyPrefix = await prefixIsEmpty(s3Prefix);
+  // TODO if the prefix is empty, we can have S3 copy everything
+  // from prod/ to the new prefix to warm things up, while we wait
+  // for gatsby build.
+
+  // First build of a PR is slow because it can't reuse cache.
+  // But we can download from prod to warm cache up.
+  const cacheWarmPrefix = emptyPrefix ? 'prod' : s3Prefix;
```
shcheklein: ideally, we should not be using the same `prod` prefix, we should be generating a new one for prod every time to make the operation atomic

.gitignore
```diff
@@ -7,3 +7,4 @@ tmp/*
 .DS_Store
 .idea
 .vscode
+404.html
```
shcheklein: why do we need to ignore it?

src/server/server.js
```diff
@@ -8,38 +10,77 @@
...
-      ],
-      source: '**'
-    }
-  ],
-  public: path.join(__dirname, '../../public'),
-  trailingSlash: false
+const { PORT = 9000, NODE_ENV } = process.env;
+
+const cacheControl = 'public, max-age=0, s-maxage=999999';
+const wreckOptions = {
```
shcheklein: do we need it as a global const?

src/server/server.js
```diff
@@ -8,38 +10,77 @@
...
+  redirects: 2,
+  timeout: 5000
 };
+const htmlType = 'text/html; charset=utf-8';
+
+let notFoundPage;
+try {
+  // in production there's no public folder
```
shcheklein: still not sure what's going on with 404

src/server/server.js
```diff
@@ -8,38 +10,77 @@
...
+    }
+  }
 });
 
-app.listen(port, e => {
+app.listen(PORT, e => {
   /* tslint:disable-next-line:no-console */
-  console.log(`Listening on http://0.0.0.0:${port}/`);
+  console.log(`Listening on http://0.0.0.0:${PORT}/`);
+  console.log(`Proxying to ${s3Url}`);
```
shcheklein: what about local run?

shcheklein: does it still deploy to S3?

shcheklein: or we don't use express middleware locally at all? when we do `yarn develop`

src/server/server.js
```diff
@@ -8,38 +10,77 @@
...
+}
+
+app.use(async (req, res) => {
+  const { pathname } = url.parse(req.url);
+
+  try {
+    await serveFile(pathname, res);
+  } catch (e) {
+    try {
+      const indexFile = (pathname + '/index.html').replace(/\/+/, '/');
```
shcheklein: why do we need this replace?

fabiosantoscode: The pathname might have a trailing slash in it.

src/server/server.js
fabiosantoscode: This is just to avoid the extra S3 request to fetch the 404 page, which we know doesn't change until the deployment dies.

We don't deploy the public folder so 404.html is moved to the root.

src/server/server.js
fabiosantoscode: yarn develop is still powered by Gatsby develop.

This is the production server, and you can run it locally with the correct environment variables. And it won't serve the public folder, instead it proxies to S3 just like production. You can also deploy locally if you need to but you need the right AWS token to do so.

I haven't really documented what variables this needs to run, so I'll do that.

.gitignore
fabiosantoscode: If you run the deploy script locally this file goes to the root and shouldn't be committed since it's generated.

scripts/deploy-with-s3.js
fabiosantoscode: Sure, sorry.

scripts/deploy-with-s3.js
fabiosantoscode: There's a bunch but in this case we need an action between the retries.

scripts/destroy-pr.js
fabiosantoscode: I can't get logs off this script to understand why it's failing or even whether it's running so I thought of doing just that over the weekend.

scripts/deploy-with-s3.js
fabiosantoscode: I'll check

scripts/deploy-with-s3.js
fabiosantoscode: Heroku doesn't generate a different environment variable for each production deployment. Plus, it doesn't provide a script to destroy production environments.

The deployment takes around 3 seconds to upload (given a low amount of files) so it's pretty atomic as it is. Plus, the cache shields us from inconsistency as long as it's warm, and gets cleared in one go.

src/server/server.js
fabiosantoscode: Not really. I'll place this right in the call.

scripts/deploy-with-s3.js
fabiosantoscode: Can do

fabiosantoscode:
@shcheklein I believe I've addressed the comments specific to code, and am now working on the lifecycle policy for the bucket.

>  Didn't get the issue with IAM.

Basically, a bad actor can send a PR which runs `env` during the build, and reads the heroku CI logs therefore gaining access to the AWS token. This can only happen if we enable automatic PRs or run preview environments carelessly.

However this cannot compromise production (separate tokens, separate ACLs), just other PRs.

src/server/server.js
shcheklein: 👍

src/server/server.js
```diff
@@ -1,45 +1,94 @@
 /* eslint-env node */
 
+'use strict';
+
 /*
- * Production server.
+ * Production server. Proxies to S3 depending on HEROKU_APP_NAME (see scripts/deploy-with-s3.js)
```
shcheklein: minor: looks >80 long to mee

fabiosantoscode:
Preview environment for this PR, served from S3, right here: https://dvc-blog-master-dmu0xfokikp14f.herokuapp.com/

PR #144, opened 2020-03-10
meta: review meta tags, keywords and favicons

fabiosantoscode:
This adds the URL metadata to each page, and also adds image width/height for facebook to be able to place a picture in the preview.

WIP: need to understand what else is missing, so I need access to the google doc about this issue, and I'm also going to run some SEO and metadata checkers I can find on the web. Open to suggestions!


shcheklein:
@fabiosantoscode shared the doc, PTAL

shcheklein:
@fabiosantoscode please, resolve the conflicts.

@fabiosantoscode I believe there are still some tags, icons, descriptions are missing from the doc I shared and from the ticket?

PR #145, opened 2020-03-11
seo: optimisation

fabiosantoscode:
## TODO

 * [x] optimal redirects to non-slash and to remove /blog/ from the start (someone's still using those links)
 * [x] Remove pagination pages from sitemap.xml
 * [x] Add `link[rel="next"]` and `link[rel="prev"] to identify next and previous pages, if any
 * [x] Change title in */page/2 and beyond to avoid duplicate titles
 * [x] Abolish `dvc` tag since it includes every post

We have few cross-article links. We might benefit from a "related articles" `<aside>`, even if it's generated from simple tag correlation. It might be helpful for users as well. If we can, we should try to link between our articles when it makes sense.


src/server/middleware/canonical-url.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = (req, res, next) => {
+  const { pathname } = url.parse(req.url);
+
+  let target = pathname;
+
+  if (target.endsWith('/') && target !== '/') {
+    target = target.slice(0, -1);
+  }
+
+  if (target.startsWith('/blog/')) {
```
shcheklein: where does it happen?

shcheklein: where did you see those links?

also, I think we should handle it similar to dvc.org - with a file with redirects and an engine that handles this.

fabiosantoscode: Found a bunch of 404's on google search console pointing to `blog.dvc.org/blog/$realposturi`

I'll integrate the redirect engine from dvc.org as found in @iAdramelk's PR. I don't remember why I didn't choose to do that in the first place.

content/blog/2020-02-04-gsoc-ideas-2020.md
```diff
@@ -126,5 +125,5 @@ worldwide community of contributors, will learn most by visiting our
 [Discord channel](https://dvc.org/chat),
 [GitHub repository](https://github.com/iterative/dvc), and
 [Forum](https://discuss.dvc.org/). We are available to discuss project proposals
-from interested students and can be reached by [email](support@dvc.org) or on
-our Discord channel.
+from interested students and can be reached by [email](mailto:support@dvc.org)
```
shcheklein: good catch!

shcheklein:
@fabiosantoscode CI failed ^^ otherwise LGTM!

PR #148, opened 2020-03-17
deploy: do not reuse all of public/

fabiosantoscode:
The idea is to only restore the thumbnails/resizes from S3, and leave the rest of the files alone. It shouldn't change the build performance.

### DIFF
--- a/scripts/deploy-with-s3.js
+++ b/scripts/deploy-with-s3.js
@@ -20,7 +20,7 @@
 
 const path = require('path');
 const { execSync } = require('child_process');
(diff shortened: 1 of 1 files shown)


fabiosantoscode:
Timings:

 * PR first build: 4:35
 * Subsequent builds: 3:20

fabiosantoscode:
So it looks like we lose between 20 and 35 seconds. My educated guess is that the CSS+JS build is to blame for a bunch of this.

I think it can be mitigated by reading the JSON file outlining the files, and pulling them from S3, but I think these timings aren't that bad.

